PR #16, opened 2016-08-26
Enforce code styling rules for Node files

santiph:
Enforce code styling rules for Node files


bisht-richa:
Looks good  :)

skazi0:
The PR seems to cover not only code style but also some more general cleanup (e.g. unused variables removal). Would be good to add it to the commit message and/or split the whitespace vs non-whitespace changes into separate commits.

PR #21, opened 2016-08-31
Fixed common layout

skazi0:
Fixed common layout:
- full height steps sidebar
- bottom margin in details part

Based on Solution1 from: http://stackoverflow.com/questions/19695784/how-can-i-make-bootstrap-columns-all-the-same-height

**Note: the height of the whole box is not fixed so it's going to change depending on the contents of the details section which might or might not be in line with the expectations. TBD.**

### DIFF
--- a/assets/app/features/upgrade/templates/upgrade7/upgrade.jade
+++ b/assets/app/features/upgrade/templates/upgrade7/upgrade.jade
@@ -1,7 +1,7 @@
 .panel.panel-default.upgrade-flow
   .panel-body
     .row
-      .col-lg-3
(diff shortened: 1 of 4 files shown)


skazi0:
This is how it looks like in firefox:
zrzut ekranu z 2016-09-01 12-07-16

And this is on a narrow screen:
zrzut ekranu z 2016-09-01 12-08-10

bisht-richa:
Looks good.

rsalevsky:
Can you make first letter from the first commit capital?

skazi0:
Updated commit message.

santiph:
Approved

PR #23, opened 2016-09-01
Features/upgrade7/upgrade admin/upgrade admin ui

skazi0:
This is a replacement of the old PR https://github.com/crowbar/crowbar-angular/pull/7 by @bisht-richa which got quite messy after multiple merges.

Some of the work in the initial commit was done from scratch based on the changes in original PR and some is a direct copy.


skazi0:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;AzhXpYt6&#x2F;99-8-s47p13-implement-upgrade-ui-part-2-3"> (8) S47P13 Implement Upgrade UI (part 2&#x2F;3)</a>

assets/app/features/upgrade/i18n/en.json
```diff
@@ -133,6 +133,14 @@
                 "form": {
                     "check-again": "Check"
                 }
+            },
+            "upgrade-admin": {
+                "name": "",
```
santiph: What is Name used for? Shouldn't it be removed?

skazi0: I'm not using it for anything but it was added like this in many places so I assumed this is some kind of workaround or something... if it's not, I'll remove it

assets/app/features/upgrade/controllers/upgrade7-upgrade-admin.controller.js
```diff
@@ -0,0 +1,66 @@
...
+
+    Upgrade7UpgradeAdminController.$inject = ['$timeout', 'upgradeUpgradeAdminFactory'];
+    // @ngInject
+    function Upgrade7UpgradeAdminController($timeout, upgradeUpgradeAdminFactory) {
+        var vm = this;
+        vm.adminUpgrade = {
+            completed: false,
+            running: false,
+            beginAdminUpgrade: beginAdminUpgrade,
+            checkAdminUpgrade: checkAdminUpgrade
```
santiph, line 23: I'm not sure we need to expose `checkAdminUpgrade` function through the vm. What do you think?

assets/app/features/upgrade/controllers/upgrade7-upgrade-admin.controller.js
```diff
@@ -0,0 +1,66 @@
...
+                    },
+                    // In case of failure
+                    function (errorResponse) {
+                        // Expose the error list to adminUpgrade object
+                        vm.adminUpgrade.errors = errorResponse.data.errors;
+                    }
+                ).finally(
+                    function () {
+                        // schedule another check if not completed yet
+                        if (!vm.adminUpgrade.completed)
```
santiph: Please, use braces on control structures, even if it's a 1 single line

assets/app/data/services/upgrade/upgrade-admin.factory.js
```diff
@@ -0,0 +1,33 @@
...
+        function getAdminUpgrade() {
+            var requestOptions = {
+                method: 'POST',
+                url: '/api/upgrade7/admin-upgrade'
+            };
+
+            return $http(requestOptions);
+        }
+
+        function getAdminUpgradeStatus() {
```
santiph, line 24: I think we should rely on the Crowbar entity to determine the status of the admin node upgrade.
Actually, that's what we need to use to determine the status of the whole Upgrade flow whenever the browser is opened.

assets/app/features/upgrade/controllers/upgrade7-upgrade-admin.controller.js
santiph: I'm surprised this wasn't caught by ESLint. 😮

assets/app/features/upgrade/controllers/upgrade7-upgrade-admin.controller.js
skazi0: We don't but I'm not sure how to test it without it :)

assets/app/data/services/upgrade/upgrade-admin.factory.js
skazi0: The whole API integration needs to be (re)checked. Both the URLs and the contents of requests and responses. I had to assume something to go forward.
Another thing is that if we are going to use one endpoint for multiple things, maybe there should be some global polling service which will check it and expose the cached contents to the page controllers...

assets/app/features/upgrade/controllers/upgrade7-upgrade-admin.controller.spec.js
```diff
@@ -0,0 +1,161 @@
...
+    });
+
+    // Verify no unexpected http call has been made
+    bard.verifyNoOutstandingHttpRequests();
+
+    it('should exist', function () {
+        should.exist(controller);
+    });
+
+    describe('nextStep function', function () {
```
santiph: I think this section could be removed, since `nextStep` function is part of (or should be soon) `https://github.com/crowbar/crowbar-angular/blob/master/assets/app/features/upgrade/upgrade7.controller.js`

assets/app/features/upgrade/controllers/upgrade7-upgrade-admin.controller.spec.js
```diff
@@ -0,0 +1,161 @@
...
+        });
+
+        it('is not running by default', function() {
+            assert.isFalse(controller.adminUpgrade.running);
+        });
+
+        describe('beginAdminUpgrade function', function () {
+            it('should be defined', function () {
+                should.exist(controller.adminUpgrade.beginAdminUpgrade);
+            });
```
santiph, line 66: `expect(controller.adminUpgrade.beginAdminUpgrade).toEqual(jasmine.any(Function));` ?

assets/app/features/upgrade/controllers/upgrade7-upgrade-admin.controller.spec.js
```diff
@@ -0,0 +1,161 @@
...
+
+            describe('when got error from api', function () {
+                it('should expose the errors through adminUpgrade.errors object', function () {
+//                    expect(controller.adminUpgrade.errors).toEqual(errorList);
+                });
+            });
+        });
+    });
+
+/*    describe('cancelUpgrade function', function () {
```
santiph: Same as `Next Step` on https://github.com/crowbar/crowbar-angular/pull/23/files#r77822851
CancelUpgrade functionality should be part of the Upgrade7 controller, so I think this section could be removed

assets/app/data/services/upgrade/upgrade-admin.factory.js
santiph: Yes, I think a new provider (`crowbarFactory`?) can be created to encapsulate this.

assets/app/features/upgrade/controllers/upgrade7-upgrade-admin.controller.spec.js
skazi0: removed

santiph:
I think I have an example of how the crowbar entity would look like that we can use, but it looks good so far...

assets/app/features/upgrade/controllers/upgrade7-upgrade-admin.controller.js
santiph: Good point.
Would you mind to add that as a comment in the code as well?

PR #24, opened 2016-09-01
lazy spinner directive

skazi0:
This directive adds a spinner which can be activated by a boolean value but will show only when active for more than specified `delay`.

To use just insert: `<suse-lazy-spinner delay="3000" active="vm.running">` into your button or any other element where the spinner is required.
Default delay (if not specified) is `2000` (2s).

If you need feedback when the spinner is actually visible, bind a scope variable to `visible` attribute.
E.g. `<suse-lazy-spinner active="vm.running" visible="vm.spinnerVisible">`.
This can be used to simplify the styling as you can change styles on the parent elements in sync with the spinner events.

In this PR there is an example of using this directive on the landing page. Note that the test backend needs to be slowed down in order to see the effect of this directive.

### DIFF
--- a/assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.js
+++ b/assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.js
@@ -1,0 +1,50 @@
+(function() {
+    'use strict';
+
+    angular
+        .module('crowbarWidgets')
+        .directive('suseLazySpinner', lazySpinner);
+
+    lazySpinner.$inject = ['$timeout'];
+
+    function lazySpinner($timeout) {
+        return {
+            restrict: 'E',
+            templateUrl: 'app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.html',
+            scope: {
+                active: '=',
+                delay: '@'
+            },
+            link: function(scope, elem) {
+                hideSpinner();
+
+                scope.$watch('active', function(newVal){
+                    newVal ? showSpinner() : hideSpinner();
+                });
+
+                var timer = null;
+
+                function showSpinner() {
+                    if (timer)
+                        return;
+                    timer = $timeout(function() { elem.css({display: ''}); }, getDelay());
+                }
+
+                function hideSpinner() {
+                    // cancel (possibly) pending timer
+                    if (timer) {
+                        $timeout.cancel(timer);
+                        timer = null;
+                    }
+
+                    elem.css({display: 'none'});
+                }
+
+                function getDelay() {
+                    var delay = parseInt(scope.delay);
+                    return isFinite(delay) ? delay : 2000;
+                }
+            }
+        };
+    }
+})();
--- a/assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.js
+++ b/assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.js
@@ -13,26 +13,28 @@
             templateUrl: 'app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.html',
             scope: {
                 active: '=',
+                visible: '=?',
                 delay: '@'
             },
             link: function(scope, elem) {
-                hideSpinner();
+                // dummy-init optional binding values
+                scope.visible = scope.visible || false;
 
                 scope.$watch('active', function(newVal) {
                     newVal ? showSpinner() : hideSpinner();
                 });
 
-                elem.on('$destroy', function() {
-                    if (timer)
-                        $timeout.cancel(timer);
-                });
+                elem.on('$destroy', hideSpinner);
 
                 var timer = null;
 
                 function showSpinner() {
                     if (timer)
                         return;
-                    timer = $timeout(function() { elem.css({display: ''}); }, getDelay());
+                    timer = $timeout(function() {
+                        elem.addClass('visible');
+                        scope.visible = true;
+                    }, getDelay());
                 }
 
                 function hideSpinner() {
@@ -42,7 +44,8 @@
                         timer = null;
                     }
 
-                    elem.css({display: 'none'});
+                    elem.removeClass('visible');
+                    scope.visible = false;
                 }
 
                 function getDelay() {
--- a/assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.js
+++ b/assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.js
@@ -18,8 +18,13 @@
             link: function(scope, elem) {
                 hideSpinner();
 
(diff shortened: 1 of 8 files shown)


assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.js
```diff
@@ -0,0 +1,58 @@
+(function() {
+    'use strict';
+
+    angular
+        .module('crowbarWidgets')
+        .directive('suseLazySpinner', lazySpinner);
```
santiph: To keep consistency, I'd use the same name for both directive's and function's name

assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.js
```diff
@@ -0,0 +1,58 @@
...
+                scope.$watch('active', function(newVal) {
+                    newVal ? showSpinner() : hideSpinner();
+                });
+
+                elem.on('$destroy', hideSpinner);
+
+                var timer = null;
+
+                function showSpinner() {
+                    if (timer)
```
santiph: Could you also use curly braces here?

assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.js
```diff
@@ -0,0 +1,58 @@
...
+                        timer = null;
+                    }
+
+                    elem.removeClass('visible');
+                    scope.visible = false;
+                }
+
+                function getDelay() {
+                    var delay = parseInt(scope.delay);
+                    return isFinite(delay) ? delay : 2000;
```
santiph: `2000` should be defined as a constant

assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.less
```diff
@@ -0,0 +1,8 @@
+/* Start= widgets/suse-lazy-spinner/suse-lazy-spinner.less */
+suse-lazy-spinner {
```
santiph, line 2: Bootstrap has a helper class to hide elements: `.hidden`. Maybe you can use such class to switch between it and `display: inline-block` as a default style.
http://getbootstrap.com/css/#helper-classes-show-hide

santiph:
Looks good so far. Please consider Bootstrap's helper classes and let me know to check again

assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.less
skazi0: done

assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.js
```diff
@@ -0,0 +1,60 @@
+(function() {
+    'use strict';
+
+    angular
+        .module('crowbarWidgets')
+        .directive('suseLazySpinner', suseLazySpinner)
+        .constant('SUSELS_DEFAULT_DELAY', 2000);
```
santiph: How about
`.constant('SUSE_LAZY_SPINNER', {DEFAULT_DELAY: 2000});`?

I'm not fully convinced a directive file is the right place for a constant to be defined, but I don't have a strong suggestion neither

skazi0: I was thinking about this too. Changed now.
As for a separate `constants` file, I think we can keep it inside the directive file as long as it's only one constant and it's used only by this directive. It can be refactored when the situation changes.

assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.spec.js
```diff
@@ -0,0 +1,125 @@
...
+        //Mock requests that are expected to be made
+        $httpBackend.expectGET('app/features/upgrade/i18n/en.json').respond({});
+        $httpBackend.flush();
+
+        scope = $rootScope.$new();
+    });
+
+    function compileDirective(tpl) {
+        var elem = angular.element(tpl);
+        var compiledElem = $compile(elem)(scope);
```
santiph: Please avoid Abbreviations.
Also, you could use

```
var elem = angular.element(tpl),
    compiledElem = $compile(elem)(scope);
```

assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.spec.js
```diff
@@ -0,0 +1,125 @@
...
+    var directiveElem,
+        scope;
+
+    beforeEach(function () {
+        bard.appModule('crowbarApp');
+
+        bard.inject('$rootScope', '$compile', '$httpBackend', '$timeout');
+
+        //Mock requests that are expected to be made
+        $httpBackend.expectGET('app/features/upgrade/i18n/en.json').respond({});
```
santiph, line 13: Are we really expecting the language file to be fetched here?

assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.spec.js
```diff
@@ -0,0 +1,125 @@
...
+
+        return compiledElem;
+    }
+
+    // Verify no unexpected http call has been made
+    bard.verifyNoOutstandingHttpRequests();
+
+    describe('directive created with default parameters', function () {
+        beforeEach(function () {
+            directiveElem = compileDirective('<suse-lazy-spinner active="testActive">');
```
santiph: Shouldn't this be instead:
`'<suse-lazy-spinner active="testActive"></suse-lazy-spinner>'`

santiph:
Good work with testing. Please, review my comments and I'll check it again

assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.spec.js
skazi0: I took this one from some example. Changed.

assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.spec.js
skazi0: From this directive's perspective we don't, but it's tested in the app context and the language file is actually fetched somewhere in the background.

assets/app/widgets/suse-lazy-spinner/suse-lazy-spinner.directive.spec.js
skazi0: Seems you're right: https://github.com/angular/angular.js/issues/1953

santiph:
Looks good. Approved

PR #25, opened 2016-09-08
increased severity of curly braces check

skazi0:
this should detect things like:
`if (true) do_someting();`


santiph:
Looks good to me

PR #27, opened 2016-09-09
switched from disabled class to ng-disabled

skazi0:
Adding class "disabled" doesn't actually disable buttons (at least in some cases). ngDisabled directive is there to handle this properly.


santiph:
Looks good to me. Approved

PR #33, opened 2016-09-13
Packaging

rsalevsky:
DO NOT MERGE!!!!!

This is just a test for the packaging.


rsalevsky:
We need to keep somehow track of this in case we need to update it. This will be the versions I use for building the app. In case we need to change something we need to adapt it and commit it.

PR #43, opened 2016-09-15
Add Buttons global styles

santiph:
Add Buttons global styles
Implementation on Backup page

### DIFF
--- a/assets/content/less/buttons/buttons.less
+++ b/assets/content/less/buttons/buttons.less
@@ -45,15 +45,15 @@
         &.spinner-visible {
             padding-left: 10px;
             padding-right: 25px;
-        }
 
-        suse-lazy-spinner {
-            i {
-                font-size: 16px;
-                width: auto;
+            suse-lazy-spinner {
+                margin-right: 9px;
+
+                i {
+                    font-size: 16px;
+                    width: auto;
+                }
             }
-
-            margin-right: 9px;
         }
     }
 
--- a/assets/content/less/buttons/buttons.less
+++ b/assets/content/less/buttons/buttons.less
@@ -15,7 +15,6 @@
     &.btn-primary:hover {
         background: #65a6de;
         border: 1px solid #5a93c4;
-        color: #ffffff;
     }
 
     // Green button
@@ -23,7 +22,6 @@
     &.btn-success:hover {
         background: #80c235;
         border: 1px solid #72ab31;
-        color: #ffffff;
     }
 
     // Grey button
--- a/assets/content/less/buttons/buttons.less
+++ b/assets/content/less/buttons/buttons.less
@@ -1,0 +1,73 @@
+// lesshint qualifyingElement: { enabled: true, allowWithClass: true, allowWithAttribute: true }
+.btn {
+    font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
+    font-size: 14px;
+    padding-left: 29px;
+    padding-right: 31px;
+
+    &:hover {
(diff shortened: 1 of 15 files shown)


assets/content/less/buttons/buttons.less
```diff
@@ -0,0 +1,73 @@
...
+    padding-right: 31px;
+
+    &:hover {
+        box-shadow: 0px 2px 4px 0px rgba(0, 0, 0, 0.23);
+    }
+
+    // Blue button
+    &.btn-primary,
+    &.btn-primary:hover {
+        background: #65a6de;
```
bisht-richa, line 16: Can use variables for colour code.

santiph: I don't think it's worthy. We have 15 colors and only 2 repetitions.

bisht-richa:
Looks good.

assets/content/less/buttons/buttons.less
```diff
@@ -0,0 +1,73 @@
...
+        border-radius: 3px;
+        color: #ffffff;
+    }
+
+    // Green button
+    &.btn-success,
+    &.btn-success:hover {
+        background: #80c235;
+        border: 1px solid #72ab31;
+        border-radius: 3px;
```
cyntss: You could Not duplicate this rule across all buttons and just define it once for all of them

assets/content/less/buttons/buttons.less
cyntss: I agree with setting variables for all colours, it makes it cleaner when a designer needs to just change some HEX in 1 file where the variables have a good naming convention also

cyntss: Also, I dont know if you are using bootstrap here, I didnt find the dependency in the bower file, but if you do, then you can simply modify the bootstrap variable value for each button. That will be much cleaner.

assets/content/less/buttons/buttons.less
santiph: Good catch. Will re-apply it

assets/content/less/buttons/buttons.less
santiph: Bootstrap is being included currently on line 27: https://github.com/crowbar/crowbar-ui/blob/master/bower.json#L27

But we're not consuming it importing the less file, but the final .min files (https://github.com/crowbar/crowbar-ui/blob/master/gulp/bowerExtract.js#L51), so I don't think variables replacement is that easy now. :/

But we could research and improve that in next sprint and track the variables replacement there, as its out of the current PR's scope.

cyntss: Oh i didnt see it in bower. Yes, in the end it is improvements to make the best out of Bootstrap, such as being able to modify variables only, include bootstrap classes inside a custom class, or use mixins, etc.

So it makes sense if you wanna do it as part of an improvement round 👍

santiph: I've created https://github.com/crowbar/crowbar-ui/issues/44 to track that work

assets/content/less/buttons/buttons.less
```diff
@@ -14,7 +15,6 @@
     &.btn-primary:hover {
         background: #65a6de;
         border: 1px solid #5a93c4;
-        border-radius: 3px;
         color: #ffffff;
```
cyntss: Are you sure you need to give color: #fff to these two buttons? because by default, Bootstrap's buttons have white text

cyntss (commented):
LGTM

PR #48, opened 2016-09-19
Fix glyphicon style

bisht-richa:
Updated glyphicon in checklist directive and update unit test file


santiph:
Looks good to me. Approved

PR #56, opened 2016-10-04
Update rsync script and options

santiph:
New `--no-sync=true/false` option added to skip remote synchronisation completely.
Add new error handlers around the configuration file and its cloud options

### DIFF
--- a/gulp/rsync.js
+++ b/gulp/rsync.js
@@ -9,8 +9,7 @@ var gulp = require('gulp'),
     syncClouds = syncCloudConfig ? syncCloudConfig.clouds : {},
     syncCloud = syncClouds ? syncClouds[syncCloudID] : undefined,
 
-    skypeRsync = 'Skypping rsync.';
-
+    skypeRsync = 'Skipping rsync.';
 
 gulp.task('rsync', function () {
     // Skip rsync option
@@ -53,6 +52,5 @@ gulp.task('rsync', function () {
                 recursive: true,
                 silent: !syncCloudConfig.options.verbose
             }));
-
     }
 });
--- a/gulp/rsync.js
+++ b/gulp/rsync.js
@@ -3,20 +3,45 @@ var gulp = require('gulp'),
     gutil = require('gulp-util'),
     rsync = require('gulp-rsync'),
     optional = require('optional'),
-    syncCloudConfig = optional('./cloud.conf.json') || {},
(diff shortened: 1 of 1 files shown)


gulp/rsync.js
```diff
@@ -3,20 +3,45 @@ var gulp = require('gulp'),
...
-    syncCloudConfig = optional('./cloud.conf.json') || {},
+    syncCloudConfig = optional('./cloud.conf.json'),
     syncCloudID = gutil.env['sync-cloud'] || 'default',
-    syncClouds = syncCloudConfig.clouds || {},
-    syncCloud = syncClouds[syncCloudID];
+    noSync = gutil.env['no-sync'] || false,
+    syncClouds = syncCloudConfig ? syncCloudConfig.clouds : {},
+    syncCloud = syncClouds ? syncClouds[syncCloudID] : undefined,
+
+    skypeRsync = 'Skypping rsync.';
```
skazi0: Should be skip/Skipping.

gulp/rsync.js
```diff
@@ -28,5 +53,6 @@ gulp.task('rsync', function () {
                 recursive: true,
                 silent: !syncCloudConfig.options.verbose
             }));
+
```
skazi0: Why was this blank line added?

skazi0 (changes requested):
I'm not sure you need this much error handling in a script which is only a part of development environment and doesn't go into the product.
IMO, if there is a template for the config file, there's no need to check in the code if the actual config matches the template.

gulp/rsync.js
santiph: good catch

gulp/rsync.js
santiph: No need. I'll remove it

santiph:
@skazi0 I started adding an option to avoid rsync at all, and then continued handling other scenarios to help to eventually debug, just for fun
btw, I've updated the changes

gulp/rsync.js
skazi0: The typo is still there in the variable name.

santiph: Should be updated now

skazi0: I'm not a native english speaker but I believe it should be `skipRsync` not `skipeRsync`.

skazi0 (approved):

PR #58, opened 2016-10-05
Service integration > Nodes Repo Checks page

santiph:
- Update repositories checks based on service response
- Update static API responses (nodejs)
- Refactor Data module
- Integrate with /api/crowbar to get the list of installed Add Ons
- Display repositories list based on available Add Ons

PR #62, opened 2016-10-12
S49P15 #11 Upgrade Admin Server: UI Story 2 - Service Integration

Itxaka:
- map the response data to the model properly
- launch the check as soon as the model is created, to check for ongoing updates
- Add proper responses to specs
- Change the button text on upgrade complete
- Extra check before setting the timeout

### DIFF
--- a/assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.js
@@ -11,9 +11,11 @@
     angular.module('crowbarApp')
         .controller('UpgradeUpgradeAdministrationServerController', UpgradeUpgradeAdministrationServerController);
 
-    UpgradeUpgradeAdministrationServerController.$inject = ['$timeout', 'crowbarFactory'];
+    UpgradeUpgradeAdministrationServerController.$inject = [
+        '$timeout', 'crowbarFactory', 'ADMIN_UPGRADE_TIMEOUT_INTERVAL'
+    ];
     // @ngInject
-    function UpgradeUpgradeAdministrationServerController($timeout, crowbarFactory) {
+    function UpgradeUpgradeAdministrationServerController($timeout, crowbarFactory, ADMIN_UPGRADE_TIMEOUT_INTERVAL) {
         var vm = this;
         vm.adminUpgrade = {
             completed: false,
@@ -29,6 +31,8 @@
         function activate() {
             // On page load we need to check to see if the upgrade is already running to set
             // the button status and the update check running
+            // TODO(itxaka): Not tested yet, tests should be done as part of card:
+            // https://trello.com/c/5fXGm1a7/45-2-27-restore-last-step
             checkAdminUpgrade();
         }
 
@@ -63,14 +67,15 @@
                     // In case of failure
                     function (errorResponse) {
                         // Expose the error list to adminUpgrade object
-                        // TODO(itxaka): Use the proper error key from the response
+                        // TODO(itxaka): Use the proper error key from the response when this is done
+                        // https://trello.com/c/chzg85j4/142-3-s49p7-error-reporting-on-crowbar-level
                         vm.adminUpgrade.errors = errorResponse.data.errors;
                     }
                 ).finally(
                     function () {
                         // schedule another check if not completed yet
                         if (!vm.adminUpgrade.completed && vm.adminUpgrade.running) {
-                            $timeout(vm.adminUpgrade.checkAdminUpgrade, 1000); // TODO: make the interval a constant
+                            $timeout(vm.adminUpgrade.checkAdminUpgrade, ADMIN_UPGRADE_TIMEOUT_INTERVAL);
                         }
                     }
                 );
--- a/assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.js
@@ -24,9 +24,13 @@
             checkAdminUpgrade: checkAdminUpgrade
         };
 
-        // On page load we need to check to see if the upgrade is already running to set
-        // the button status and the update check running
-        checkAdminUpgrade();
+        activate();
+
+        function activate() {
+            // On page load we need to check to see if the upgrade is already running to set
+            // the button status and the update check running
+            checkAdminUpgrade();
+        }
 
         function beginAdminUpgrade() {
             vm.adminUpgrade.running = true;
@@ -67,10 +71,6 @@
                         // schedule another check if not completed yet
                         if (!vm.adminUpgrade.completed && vm.adminUpgrade.running) {
                             $timeout(vm.adminUpgrade.checkAdminUpgrade, 1000); // TODO: make the interval a constant
-                        } else if (vm.adminUpgrade.completed && !vm.adminUpgrade.running) {
-                            // TODO(itxaka): can translations be done from here instead so
-                            // we can update the button text for better UX?
-                            vm.finish_text = 'Upgrade done!'
                         }
                     }
                 );
--- a/assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.js
@@ -24,6 +24,10 @@
             checkAdminUpgrade: checkAdminUpgrade
         };
 
+        // On page load we need to check to see if the upgrade is already running to set
+        // the button status and the update check running
+        checkAdminUpgrade();
+
         function beginAdminUpgrade() {
             vm.adminUpgrade.running = true;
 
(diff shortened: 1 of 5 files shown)


Itxaka:
Criteria was:
A mechanism to start the admin server upgrade, and to check when it is finished is implemented.
Expected temporary API downtime (e.g. when the admin server is rebooting) is handled properly (i.e. the error seen by the UI during this time window is not treated as an ordinary error)

In my tests I have found:
- Start the upgrade -> Already done and in place
- Check when its finished -> was almost done and only required a couple of canges to the reponse workflow
- Temporary downtime -> When the node is down the service already keeps checking. There is nothing in the UI to indicate that there is an error anywhere.

Problems with this:
- If you launch the upgrade and refresh the page, the state is lost. I added a first check on model initialisation so if the upgrade is ongoing it will disable the button and start the periodic check for status.
-  Im still not sure what kind of errors the API can return nor if it can even return anything (returns 422 if you try to start the upgrade and the upgrade is ongoing but that cant happen anymore)
- Next button is enabled all the time, nor there is anything in the model to flag the current step as finished so the button is enabled on step completition.

Itxaka:

Itxaka:
Oh yeah, on testing, due to the controller being created beforeEach at the top, I am unable to override the `initialResponse` called on controller creation (the first call to `checkAdminUpgrade()`).

I would like to be able to modify this for a subset of tests in order to initialise the model with different initial statuses. Other than creating the controller per `describe` section, Im not sure how to do it. Any thoughts?

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.js
```diff
@@ -24,6 +24,10 @@
             checkAdminUpgrade: checkAdminUpgrade
         };
 
+        // On page load we need to check to see if the upgrade is already running to set
+        // the button status and the update check running
+        checkAdminUpgrade();
```
santiph: Could you please use an activate() function here?
https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#controller-activation-promises
Also, to keep consistency with the Nodes Repo checks controller:
https://github.com/crowbar/crowbar-ui/pull/58/files#diff-2230c16354311fde135c40b1f3902a44R72

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.js
```diff
@@ -48,21 +52,25 @@
...
                         // schedule another check if not completed yet
-                        if (!vm.adminUpgrade.completed) {
+                        if (!vm.adminUpgrade.completed && vm.adminUpgrade.running) {
                             $timeout(vm.adminUpgrade.checkAdminUpgrade, 1000); // TODO: make the interval a constant
-                        } else {
-                            vm.adminUpgrade.running = false;
+                        } else if (vm.adminUpgrade.completed && !vm.adminUpgrade.running) {
+                            // TODO(itxaka): can translations be done from here instead so
+                            // we can update the button text for better UX?
+                            vm.finish_text = 'Upgrade done!'
```
santiph: The translation service can be injected into controllers and used to send text to the view (see https://angular-translate.github.io/docs/#/api/pascalprecht.translate.$translate), but I'd rather use such case only if no other option is possible.
I'd prefer using the VM to expose the state to the view and let it decide what to show, where and when.

Another possibility is to let the controller build and update language keys to be exposed to the view, and use the translations there.

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.js
```diff
@@ -48,21 +52,25 @@
...
                     function (errorResponse) {
                         // Expose the error list to adminUpgrade object
+                        // TODO(itxaka): Use the proper error key from the response
                         vm.adminUpgrade.errors = errorResponse.data.errors;
                     }
                 ).finally(
                     function () {
                         // schedule another check if not completed yet
-                        if (!vm.adminUpgrade.completed) {
+                        if (!vm.adminUpgrade.completed && vm.adminUpgrade.running) {
```
santiph, line 77: Now that I see both in a single condition, doesn't each of those attributes suggest the opposite value in the other already? Sounds a bit redundant to me.
What do you think @skazi0 ?

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.spec.js
```diff
@@ -2,12 +2,36 @@
...
     var controller,
         completedUpgradeResponse = {
-            data: { completed: true }
+            data: {
+                version: '3.0',
+                addons: [],
+                upgrade: {
+                    upgrading: false,
+                    success: true,
+                    failed: false}
```
santiph: Please, closing brace in a new line

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.spec.js
```diff
@@ -2,12 +2,36 @@
...
         },
         incompleteUpgradeResponse = {
-            data: { completed: false }
+            data: {
+                version: '3.0',
+                addons: [],
+                upgrade: {
+                    upgrading: true,
+                    success: false,
+                    failed: false}
```
santiph: same here

assets/app/features/upgrade/templates/upgrade-administration-server-page.jade
```diff
@@ -10,4 +10,4 @@ button.btn.btn-success.center-block(
 )
     suse-lazy-spinner(delay="2000", active="upgradeAdminUpgradeVm.adminUpgrade.running",
         visible="upgradeAdminUpgradeVm.adminUpgrade.spinnerVisible")
-    span(translate='') upgrade.steps.upgrade-admin.form.upgrade-admin
+    span(translate='', ng-bind="upgradeAdminUpgradeVm.finish_text") upgrade.steps.upgrade-admin.form.upgrade-admin
```
santiph: Doesn't ngBind generate an unnecessary watcher?

santiph:
@Itxaka Don't worry on the errors the API could eventually return and/or the case the API doesn't come back at all. It's not part of the scope of this card and we have a specific one for Error Handling for this page: https://trello.com/c/kxvh6yP2/18-upgrade-admin-server-error-handling-story-1 (Still to be refined, tho).
Same thing with Next button being enabled/disabled. That should be cover as part of the navigation bar user story: https://trello.com/c/LNr710jO/104-3-s49p20-6-layout-navigation-bar

There is a static service in place returning fake data for the UI that you could use to simulate the downtime on the Admin node and see how angular would respond to that: https://github.com/crowbar/crowbar-ui/blob/master/routes/api/crowbar/upgrade.js
Would you mind to update the static responses also to match the documentation?

Finally, I left a couple of comments on the review... Looks good so far

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.js
Itxaka: This is done to avoid the timeout from firing on page load, as we are now calling the checkAdminUpgrade on controller creation, it wiould always trigger it. Instead we check both values so it only triggers it when the upgrade is ongoing.

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.js
Itxaka: Will do, thanks!

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.spec.js
Itxaka: will do

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.spec.js
Itxaka: will do

assets/app/features/upgrade/templates/upgrade-administration-server-page.jade
Itxaka: This was only done as a test for the text change on upgrade finish and for the translation question. It will be removed in the next comimt :)

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.js
skazi0: They are not direct opposite of each other. `completed` means a complete run was finished (success or fail). `running` means something is currently running. Before anything is started it's neither `completed` nor `running`.

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.js
```diff
@@ -24,6 +24,14 @@
             checkAdminUpgrade: checkAdminUpgrade
         };
 
+        activate();
+
+        function activate() {
+            // On page load we need to check to see if the upgrade is already running to set
+            // the button status and the update check running
+            checkAdminUpgrade();
```
santiph, line 36: We need to test this functionality, somehow. I'm still not sure about the proper approach for this that we should follow here (and in the following steps). But is not part of the scope of the current card we've assigned this sprint.
It's part of Restore last step one, as detailed in https://trello.com/c/5fXGm1a7/45-2-27-restore-last-step (Has not been properly refined, estimated, prioritised nor committed yet. ;) )

santiph: Don't remove it, just add a comment saying it will be properly implemented and tested as part of that card

santiph:
Looks good to me. Please address the comment regarding checking the current status and consider this approved

santiph:
Approved

PR #65, opened 2016-10-13
Whitespace cleanup

skazi0:
Removed some trailing white spaces and unneeded blank lines.


santiph:
Thanks for this cleanup. Would you mind to enforce our lints configuration as well? That way we could have this prevented in the future

skazi0:
I enabled `eslint` linter for trailing whitespace. The `lesshint` linter was enabled by default but it only generates warnings/errors and there seem to be no option to force failing on error.

PR #68, opened 2016-10-17
Add tests for PR #62

Itxaka:
As we are modifying the controller status on initialization to be able to track the ongoing upgrade in case it was started before the controller creation, we need to test those initial values.
Moves the controller creation from the top file to some of the tests and/or sections to be able to modify the initial response on some of the tests.
- Move the controller creation to a function to have control over initial values
- Add tests for status update on controller initialization

### DIFF
--- a/assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.spec.js
+++ b/assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.spec.js
@@ -39,7 +39,7 @@ describe('Upgrade Flow - Upgrade Administration Server Controller', function () 
         },
         mockedTimeout;
 
-    beforeEach(function() {
+    function createcontroller() {
         //Setup the module and dependencies to be used.
         bard.appModule('crowbarApp');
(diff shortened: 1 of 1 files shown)


Itxaka:
Could also probably do a beforeEach to inject the controller inside a describe that captures all tests minus the 3 new ones and move those to a different describe in order to not have to repeat the `controller = createController()` part so many times, but not sure if we rather have the tests properly organized than having to repeat a line several times?

Itxaka:

santiph:
You don't need to assign `controller = createController()`, it's already happening on line 53.

Also, if required, you can separate tests as we did for navigation test in landing page: https://github.com/crowbar/crowbar-ui/blob/master/assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js#L203-L235

Finally, what you're trying to do really looks like a `beforeEach()`. There should be a better approach somewhere...

Itxaka:
Closed in favor of https://github.com/crowbar/crowbar-ui/pull/69

PR #69, opened 2016-10-17
Add new tests for  controller initialization

Itxaka:
As we are modifying the controller status on initialization to be able to track the ongoing upgrade in case it was started before the controller creation, we need to test those initial values.

### DIFF
--- a/assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.spec.js
+++ b/assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.spec.js
@@ -52,6 +52,13 @@ describe('Upgrade Flow - Upgrade Administration Server Controller', function () 
         //Create the controller
         controller = $controller('UpgradeUpgradeAdministrationServerController', { '$timeout': mockedTimeout });
 
+        // reset the initialResponse to the default values
+        initialResponse.upgrade = {
+            upgrading: false,
+            success: false,
+            failed: false
+        };
+
         //Mock requests that are expected to be made
         $httpBackend.expectGET('app/features/upgrade/i18n/en.json').respond({});
         $httpBackend.expectGET('/api/crowbar/upgrade').respond(initialResponse);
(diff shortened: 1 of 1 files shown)


Itxaka:

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.spec.js
```diff
@@ -65,6 +65,44 @@ describe('Upgrade Flow - Upgrade Administration Server Controller', function ()
...
+        });
+
+        it('should set completed to true if upgrade is completed', function () {
+            initialResponse.upgrade.success = true;
+            controller = $controller('UpgradeUpgradeAdministrationServerController');
+            $httpBackend.expectGET('/api/crowbar/upgrade').respond(initialResponse);
+            $httpBackend.flush();
+            assert.isTrue(controller.adminUpgrade.completed);
+            assert.isFalse(controller.adminUpgrade.running);
+        })
```
santiph: semicolon

Itxaka: done!

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.spec.js
```diff
@@ -65,6 +65,44 @@ describe('Upgrade Flow - Upgrade Administration Server Controller', function ()
         should.exist(controller);
     });
 
+    describe('on controller creation', function () {
+        afterEach(function () {
```
santiph: Could this be a beforeEach() in a specific describe section for the last two cases?

Itxaka: if the `initialResponse` is not reset after being modified on a test, it can affect other tests by reporting incorrect statuses due to the last modification, so we need to reset the `initialResponse` after each of those tests.

Also, the first test is not necessary, I just saw that those vars are already tested below. Ill remove it.

skazi0: It could be reset in the top level `beforeEach()` and it should be clean for every test no matter where it's executed. See http://coderdiaries.com/2014/05/30/jasminejs-execution-sequence-whith-nested-describes/

santiph (approved):
Looks good to me. Approved

PR #72, opened 2016-10-18
Updated service integration for the landing page.

skazi0:


skazi0:
[(3) S49P12 #2 Landing Page: UI Story 2 - Service Integration](https://trello.com/c/o6ZgjeOh/145-3-s49p12-2-landing-page-ui-story-2-service-integration)

assets/app/features/upgrade/upgrade.constants.js
```diff
@@ -8,5 +8,9 @@
             'openstack': ['SUSE-OpenStack-Cloud-7-Pool', 'SUSE-OpenStack-Cloud-7-Updates'],
             'ceph': ['SUSE-Enterprise-Storage-4-Pool', 'SUSE-Enterprise-Storage-4-Updates']
         })
+        .constant('ADDONS_PRECHECK_MAP', {
+            'ha': ['clusters_healthy'],
```
santiph, line 12: Could we use the value here directly?
`'ha': 'clusters_healthy'`

assets/app/features/upgrade/controllers/upgrade-landing.controller.js
```diff
@@ -23,26 +40,36 @@
...
 
+        activate();
+
+        /**
+         * Check installed add-ons on page load.
+         */
+        function activate() {
+            crowbarFactory.getEntity().then(function (response) {
+                _.forEach(response.data.addons, function (addon) {
+                    _.forEach(ADDONS_PRECHECK_MAP[addon], function (precheck) {
```
santiph, line 67: If we convert the key:array map to a single key:string one, we could get rid of this forEach

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
```diff
@@ -91,6 +111,29 @@ describe('Upgrade Landing Controller', function() {
                     assert.isFalse(value.status);
                 });
             });
+
+            it('should contain checks for addons', function () {
+                expect(controller.prechecks.checks).toEqual(
+                    jasmine.objectContaining({
+                        'clusters_healthy': jasmine.anything(),
```
santiph: Could `jasmine.anything()` be avoided and test an object with status and label is still present?

assets/app/features/upgrade/controllers/upgrade-landing.controller.js
```diff
@@ -65,32 +92,31 @@
                 .getPreliminaryChecks()
                 .then(
                     //Success handler. Al precheck passed successfully:
-                    function(prechecksResponse) {
+                    function(response) {
 
-                        _.forEach(prechecksResponse.data, function(value, key) {
-                            vm.prechecks.checks[key].status = value;
+                        _.forEach(response.data, function(value, key) {
+                            vm.prechecks.checks[key].status = value.passed || !value.required;
```
santiph, line 108: Why `value.passed || !value.required`? Shouldn't we distinguish both cases?

assets/app/features/upgrade/upgrade.constants.js
skazi0: We could as long as there's only one precheck for each addon.

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
skazi0: We would need to add another library for such specific matching (e.g. https://github.com/JamieMason/Jasmine-Matchers)

assets/app/features/upgrade/controllers/upgrade-landing.controller.js
skazi0: Not in the old design. Here the best we can do (without modifying the checklist directive) is treat the non-mandatory checks as passed even if they're not. Otherwise these would block the upgrade.

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
skazi0: I added some more specific check but there's no builtin matcher for booleans so this would require adding a library or creating custom matcher.

skazi0: I checked again and contrary to what I found before, `jasmine.any(Boolean)` works just fine as well.

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
```diff
@@ -91,6 +110,33 @@ describe('Upgrade Landing Controller', function() {
...
             });
+
+            it('should contain checks for addons', function () {
+                expect(controller.prechecks.checks).toEqual(
+                    jasmine.objectContaining({
+                        'clusters_healthy': jasmine.objectContaining(
+                            { status: jasmine.any(Boolean), label: jasmine.any(String) }
+                        ),
+                        'ceph_healthy': jasmine.objectContaining(
+                            { status: jasmine.any(Boolean), label: jasmine.any(String) }
```
santiph: If the language keys match checks' keys, we could also assert we're using the right language key for each check.
Take a look at https://github.com/crowbar/crowbar-ui/blob/master/assets/app/features/upgrade/controllers/upgrade-nodes-repositories-checks.controller.spec.js#L306-L346

santiph:
Please consider to update/assert the language keys being used as labels.
Other than that, looks very good. Approved

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
skazi0: If you want this kind of check here, the whole label can be hardcoded in the expectation but I think the main point of this check is that the keys are as they should.

santiph: I think it would be a nice to have so we don't mess around language keys without noticing (making unit tests fail).
But I don't consider it's a blocker for these changes to be merged, so up to you

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
```diff
@@ -203,16 +251,33 @@ describe('Upgrade Landing Controller', function() {
...
+        bard.mockService(crowbarFactory, {
+            getEntity: $q.when(entityResponseWithoutAddons)
+        });
+
         controller = $controller('UpgradeLandingController');
 
+        //Mock requests that are expected to be made
+        $httpBackend.expectGET('app/features/upgrade/i18n/en.json').respond({});
+        $httpBackend.flush();
     });
```
Itxaka, line 282: Missing `bard.verifyNoOutstandingHttpRequests();`

routes/api/upgrade/prechecks.js
```diff
@@ -9,10 +9,11 @@ router.get('/', function(req, res) {
         res.status(500).json({'errors': errors});
     } else {
         res.status(200).json({
```
Itxaka, line 13: Can we get a rotating response that returns passed for half of the checks on the first try, then returns true for all? its much nicer to fully test the response locally.

assets/app/features/upgrade/i18n/en.json
```diff
@@ -23,6 +23,7 @@
                         "updates_installed": "Maintenance Updates to SLES SP1",
                         "network_sanity": "Cloud Health",
                         "high_availability": "High Availability Health",
+                        "storage": "Storage Health",
                         "free_node_available": "Free node available"
```
Itxaka, line 27: This one is not used anymore no?

assets/app/features/upgrade/upgrade.constants.js
Itxaka: Seems sensible to me to have an array in case we introduce further checks per addon in the future.

Itxaka (changes requested):
Some nitpicks here and there, all minor stuff.

Although testing it locally I cannot go through the first page, after clicking "Begin upgrade" I get an error:

```
POST /api/upgrade/prepare 404 2.358 ms - 1103
Error: Not Found
    at /home/itxaka/projects/crowbar-ui/app.js:44:15
    at Layer.handle [as handle_request] (/home/itxaka/projects/crowbar-ui/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/home/itxaka/projects/crowbar-ui/node_modules/express/lib/router/index.js:312:13)
    at /home/itxaka/projects/crowbar-ui/node_modules/express/lib/router/index.js:280:7
    at Function.process_params (/home/itxaka/projects/crowbar-ui/node_modules/express/lib/router/index.js:330:12)
    at next (/home/itxaka/projects/crowbar-ui/node_modules/express/lib/router/index.js:271:10)
    at /home/itxaka/projects/crowbar-ui/node_modules/express/lib/router/index.js:618:15
    at next (/home/itxaka/projects/crowbar-ui/node_modules/express/lib/router/index.js:256:14)
    at Function.handle (/home/itxaka/projects/crowbar-ui/node_modules/express/lib/router/index.js:176:3)
    at router (/home/itxaka/projects/crowbar-ui/node_modules/express/lib/router/index.js:46:12)
```

Im not sure why suddenly its doing a POST request when clicking on "Begin upgrade"? On master there is no request done, so it looks like its going out of the routing?? No idea tbh, may be a local issue? I cleaned up the public folder and run gulp, but the output is the same.

assets/app/features/upgrade/i18n/en.json
skazi0: It is. Why do you think it isn't?

Itxaka: Wrong line, was meant for high_availability but I can see that it has moved to optionalPreChecks so ignore this comment :+1:

routes/api/upgrade/prechecks.js
santiph: Is this part of this card's scope or is it error handling?

skazi0: This is meant to simplify testing this. Failed checks are not errors in this context.

Itxaka (approved):
Great stuff!

PR #73, opened 2016-10-18
Run extra tests to check file header failure on backup file download

Itxaka:
We were not checking if the tests were passing in case of the file headers not being correct. This adds the same tests as when the backup success but passing empty headers as to check for proper download process.

### DIFF
--- a/assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
+++ b/assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
@@ -114,14 +114,63 @@ describe('Upgrade Flow - Backup Controller', function() {
             });
 
             describe('when executed', function () {
-
-                describe('on success', function () {
+                describe('on success with wrong file headers', function () {
                     var downloadBackupMock;
 
                     beforeEach(function () {
 
                         // Mock the click() method of a fake downloadBackup element
                         downloadBackupMock = $document[0].createElement('a')
+                        spyOn(downloadBackupMock, 'click');
+
+                        // Mock the createElement() method of $document[0]
+                        spyOn($document[0], 'createElement')
+                            .and.returnValue(downloadBackupMock);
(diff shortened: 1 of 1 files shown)


skazi0:
Would it be possible to check if the default filename was used as a fallback?

Itxaka:
Yup, it is possible with some regex magic against the downloadBackupMock. Will add tomorrow!

assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
```diff
@@ -114,7 +114,51 @@ describe('Upgrade Flow - Backup Controller', function() {
...
+                        });
+
+                        // Run the backup get function
+                        controller.backup.download(42);
+                        $rootScope.$digest();
+                    });
 
+                    it('click() method should have been triggered to download the backup', function () {
+                        expect($document[0].createElement).toHaveBeenCalledWith('a');
+                        expect(downloadBackupMock.click).toHaveBeenCalled();
```
santiph, line 147: Could you also add `expect(downloadBackupMock.click.calls.count()).toBe(1);` so we assure only once this is being called?

santiph: or `assert.isTrue(downloadBackupMock.click.calledOnce);`

santiph:
Minor comment. Please address it and consider this PR approved

assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
Itxaka: Sounds good!, will do

Itxaka:
Updated to fix the comments and add the extra checks for the filename

santiph (approved):
Looks good to me. Approved

Issue #32, opened 2016-09-13
[backup] backup download doesn't work in Firefox


skazi0:
The request is sent to the API but the "save as" part doesn't seem to work properly.

santiph:
I think this has been already fixed on #82 
Could we close this issue? @skazi0 @Itxaka

Itxaka:
Should be fixed!

PR #76, opened 2016-10-19
Change the API route for admin repocheck

Itxaka:
Due to the movement of the admin repocheck to the upgrade namespace,
we need to change the route we use to call the api

Depends on https://github.com/crowbar/crowbar-core/pull/739
- Move the call to the admin repo check to the proper namespace
- Move the fake API route to the proper namespace


Itxaka:
[(3) S49P19 #32 Admin Server Repo Checks: Service Integration](https://trello.com/c/oGniN5Lf/149-3-s49p19-32-admin-server-repo-checks-service-integration)

assets/app/data/crowbar/services/crowbar.factory.js
```diff
@@ -41,7 +41,7 @@
```
santiph: (for some reason I cannot add this comment on line 12)
This method (`getRepositoriesChecks`) needs to be moved from crowbar factory to upgrade factory, as they're intended to be a reflection of the API entities.

santiph (changes requested):
Crowbar Factory method and its test needs to be moved to the Upgrade Factory.

assets/app/data/crowbar/services/crowbar.factory.js
Itxaka: out of scope for this patch or wrong PR?

Itxaka:
Ready to review as dependant patch has been merged!

assets/app/data/crowbar/services/crowbar.factory.js
santiph: I don't think so.
Repository checks for the admin server now is part of the upgrade API entity, not crowbar.
Therefore, in order to keep consistency, I think we should reflect that by moving the `getRepositoriesChecks` method from crowbar.factory.js into upgrade.factory.js

Itxaka: Ahhh I see now! Indeed you are rigth, this should be moved into the appropiate place now that has changed places in the API

santiph (approved):

santiph:
Looks good to me now. :) Thanks!

PR #82, opened 2016-10-27
Use angular-file-saver for the download files

Itxaka:
As pointed by @skazi0 angular-file-saver makes the downloading stuff pretty easy and it works cross-browser, fixing https://github.com/crowbar/crowbar-ui/issues/32 and probably any other issue with IE browsers.

This is only a run test, tests doesn't pass and there is no new tests done for the different functionality but its so @santiph can weight on what he thinks about this and if we should go forward with this or not.

### DIFF
--- a/assets/app/features/upgrade/controllers/upgrade-backup.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-backup.controller.js
@@ -81,14 +81,9 @@ function() {
                         var headers = response.headers(),
 
                             // Get the filename from the headers or default to "crowbarBackup"
-                            filename = extractFilename(headers) || 'crowbarBackup',
+                            filename = extractFilename(headers) || 'crowbarBackup';
 
-                            // Determine the content type from the header
-                            contentType = headers['content-type'],
-
-                            backupBlob = new Blob([response.data], {type: contentType});
-
-                        FileSaver.saveAs(backupBlob, filename)
+                        FileSaver.saveAs(response.data, filename)
                     },
                     // In case of download error
                     function (errorResponse) {
--- a/assets/app/features/upgrade/controllers/upgrade-backup.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-backup.controller.js
@@ -1,4 +1,4 @@
-(/*global Blob URL */
+(/*global */
 function() {
     'use strict';
 
@@ -18,7 +18,9 @@ function() {
         'crowbarUtilsFactory',
         '$document',
         'upgradeStepsFactory',
-        'upgradeFactory'
+        'upgradeFactory',
+        'FileSaver',
+        'Blob'
     ];
     // @ngInject
     function UpgradeBackupController(
@@ -27,7 +29,9 @@ function() {
         crowbarUtilsFactory,
         $document,
         upgradeStepsFactory,
-        upgradeFactory
+        upgradeFactory,
+        FileSaver,
+        Blob
     ) {
         var vm = this;
         vm.backup = {
@@ -82,18 +86,9 @@ function() {
                             // Determine the content type from the header
                             contentType = headers['content-type'],
 
-                            backupBlob = new Blob([response.data], {type: contentType}),
-                            backupObjectUrl = URL.createObjectURL(backupBlob),
+                            backupBlob = new Blob([response.data], {type: contentType});
 
-                            // Create download a DOM element
-                            backupElement = $document[0].createElement('a');
-
-                        // Set anchor properties
-                        backupElement.href = backupObjectUrl;
-                        backupElement.download = filename;
-
-                        // Trigger the download
-                        backupElement.click();
+                        FileSaver.saveAs(backupBlob, filename)
                     },
                     // In case of download error
                     function (errorResponse) {
--- a/assets/app/features/upgrade/controllers/upgrade-backup.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-backup.controller.js
@@ -30,8 +30,7 @@ function() {
         $document,
         upgradeStepsFactory,
         upgradeFactory,
-        FileSaver,
-        Blob
+        FileSaver
     ) {
         var vm = this;
         vm.backup = {
--- a/assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
+++ b/assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
@@ -10,9 +10,10 @@ describe('Upgrade Flow - Backup Controller', function() {
             data: { id: 42 }
         },
         mockedDownloadFile = '--Mock Backup File--',
+        mockedFileName = 'S33z8qFX.jpg.zip',
         mockedDownloadResponseHeaders = {
             'connection': 'keep-alive',
-            'content-disposition': 'attachment; filename=S33z8qFX.jpg.zip',
+            'content-disposition': 'attachment; filename=' + mockedFileName,
             'content-type': 'application/zip',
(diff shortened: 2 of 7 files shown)


assets/app/features/upgrade/controllers/upgrade-backup.controller.js
```diff
@@ -68,18 +75,9 @@ function() {
                             // Determine the content type from the header
                             contentType = headers['content-type'],
 
-                            backupBlob = new Blob([response.data], {type: contentType}),
-                            backupObjectUrl = URL.createObjectURL(backupBlob),
+                            backupBlob = new Blob([response.data], {type: contentType});
```
skazi0: Depending on the portability requirements (i.e. how old browsers need to be supported), we can make it even simpler: if you set `responseType: 'blob'` in `crowbar-backup.factory.js` you will get `response.data` already in blob format and you can drop the whole manual `Blob` object creation.

`Blob` has pretty decent support in modern desktop browsers: https://developer.mozilla.org/en-US/docs/Web/API/Blob#Browser_compatibility

Itxaka: I thougth that to trigger the download of zip files and binary files in general you were forced to set the `arraybuffer` responseType?

skazi0: Hmmm... haven't heard about such requirement. I tested `responseType: blob` in current FF and Chrome and it works correctly.

Itxaka:
@skazi0 ready for a second round of checks, made the tests work with the new fileSaver, used blob as you suggested and removed non used code.

assets/app/features/upgrade/controllers/upgrade-backup.controller.js
```diff
@@ -12,9 +12,16 @@ function() {
...
         .controller('UpgradeBackupController', UpgradeBackupController);
 
-    UpgradeBackupController.$inject = ['$translate', '$state', 'crowbarBackupFactory', '$document'];
+    UpgradeBackupController.$inject = [
+        '$translate',
+        '$state',
+        'crowbarBackupFactory',
+        '$document',
+        'FileSaver',
+        'Blob'
```
skazi0, line 23: is this still needed?

assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
```diff
@@ -161,21 +148,15 @@ describe('Upgrade Flow - Backup Controller', function() {
                     });
 
                     it('uses the default name for the file', function () {
-                        expect(downloadBackupMock.download).toEqual('crowbarBackup');
+                        expect(FileSaver.saveAs).toHaveBeenCalledWith(
+                          mockedDownloadResponse.data, 'crowbarBackup'
```
skazi0, line 161: Since this test is about the filename, I'd put jasmine.anything() as the first argument. The data can be a separate test.

assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
```diff
@@ -210,7 +186,9 @@ describe('Upgrade Flow - Backup Controller', function() {
                     });
 
                     it('does not use the default name for the file', function () {
-                        expect(downloadBackupMock.download).not.toEqual('crowbarBackup');
+                        expect(FileSaver.saveAs).not.toHaveBeenCalledWith(
+                          mockedDownloadResponse.data, 'crowbarBackup'
```
skazi0: You can use this place to do some basic blackbox testing of `extractFilename()`, i.e. expect the actual filename from headers here.

skazi0 (commented):
small nits in the tests. besides that, looks good.

skazi0:
[(1) S50P12 (S49P13) #8 Backup Admin Server: Service Integration](https://trello.com/c/hw0n8t1G/129-1-s50p12-s49p13-8-backup-admin-server-service-integration)

skazi0 (approved):

matelakat (approved):

skazi0:
@rsalevsky this adds an npm dependency: https://github.com/crowbar/crowbar-ui/pull/82/files#diff-0a08a7565aba4405282251491979bb6bR27 please confirm it's ok from the packaging point of view

assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
```diff
@@ -161,21 +149,15 @@ describe('Upgrade Flow - Backup Controller', function() {
                     });
 
                     it('uses the default name for the file', function () {
-                        expect(downloadBackupMock.download).toEqual('crowbarBackup');
+                        expect(FileSaver.saveAs).toHaveBeenCalledWith(
+                          jasmine.anything(), 'crowbarBackup'
```
santiph: 4 spaces please. Hasn't this been caught by travis?

assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
```diff
@@ -192,11 +174,6 @@ describe('Upgrade Flow - Backup Controller', function() {
                         $rootScope.$digest();
                     });
 
-                    it('click() method should have been triggered to download the backup', function () {
```
santiph, line 203: Why removed?

assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
```diff
@@ -209,8 +186,16 @@ describe('Upgrade Flow - Backup Controller', function() {
...
-                        expect(downloadBackupMock.download).not.toEqual('crowbarBackup');
+                    it('uses the name from the header', function () {
+                        expect(FileSaver.saveAs).toHaveBeenCalledWith(
+                          jasmine.anything(), mockedFileName
+                        );
+                    })
+
+                    it('calls saveAs with data received from the service', function () {
+                        expect(FileSaver.saveAs).toHaveBeenCalledWith(
+                          mockedDownloadResponse.data, jasmine.anything()
```
santiph: could this `anything()` be more specific?

assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
santiph: Also, could `anything()` be more specific?

santiph (changes requested):
Please, also include this dependency in the core module, so it can be reused on future projects
`assets/app/core/crowbar-core.module.js`

rsalevsky:
I have to update the package after merging this. So I would suggest once @santiph approved I will merge it and update the package.

assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
skazi0: This is the whole beauty of using library instead of self-crafted direct solution: you don't need to test the internals of the library, just make sure that library functions are called (`expect(FileSaver.saveAs).toHaveBeenCalledWith(...)`)

Itxaka:
Thank you @skazi0 for updating the PR! :+1: 

@santiph can we merge this?

assets/app/features/upgrade/controllers/upgrade-backup.controller.spec.js
santiph: You're right. I see it now

santiph:
@Itxaka Looks good so far. Please rebase it and I'll do a final review.

Itxaka:
rebased!

Itxaka:
Fixes issues that came from rebasing. Tested on both firefox+chrome. Looking good!

santiph (approved):

PR #84, opened 2016-11-11
adapt to the backup api change

MaximilianMeister:
the backup api in the upgrade case now requires an additional post
parameter

### depends on
https://github.com/crowbar/crowbar-core/pull/770
and https://github.com/crowbar/crowbar-core/pull/781

### related change in crowbar-client
https://github.com/crowbar/crowbar-client/pull/120


MaximilianMeister:
@skazi0 @santiph could you please have a look? i am not sure if the header injection works like that, and maybe there are some other bits missing?

assets/app/data/crowbar/services/crowbar-backup.factory.js
```diff
@@ -4,9 +4,9 @@
         .module('suseData.crowbar')
         .factory('crowbarBackupFactory', crowbarBackupFactory);
 
-    crowbarBackupFactory.$inject = ['$q', '$http', '$filter', 'COMMON_API_V1_HEADERS'];
```
skazi0, line 7: you should replace V1 headers injection with V2 if V1 is not used anymore.

assets/app/data/crowbar/services/crowbar-backup.factory.js
```diff
@@ -23,9 +23,9 @@
 
             var requestOptions = {
                 method: 'POST',
-                url: '/utils/backups',
```
skazi0, line 26: the data services should be split by the first url component (after `api`) so this whole function should be moved to `upgradeFactory`.

skazi0 (dismissed):

skazi0 (approved):
lgtm

assets/app/data/crowbar/services/crowbar-backup.factory.spec.js
```diff
@@ -36,7 +36,7 @@ describe('Crowbar Backup Factory', function () {
         describe('when create method is executed', function () {
 
             beforeEach(function () {
-                $httpBackend.expect('POST', '/utils/backups', undefined)
+                $httpBackend.expect('POST', '/api/upgrade/adminbackup', undefined)
```
santiph: Please, test Headers V2 are being sent as well
@see https://docs.angularjs.org/api/ngMock/service/$httpBackend
`expect(method, url, [data], [headers], [keys]);` section for method definition, and `Unit testing with mock $httpBackend` for an example.

santiph (changes requested):
Overall good changes. Since we changed it twice lately, I'd like to enforce API V2 headers are being sent.

assets/app/data/crowbar/services/crowbar-backup.factory.spec.js
```diff
@@ -36,7 +36,8 @@ describe('Crowbar Backup Factory', function () {
         describe('when create method is executed', function () {
 
             beforeEach(function () {
-                $httpBackend.expect('POST', '/utils/backups', undefined)
+                $httpBackend.expect('POST', '/api/upgrade/adminbackup', undefined, function (headers) {
+                    return headers.Accept = COMMON_API_V2_HEADERS.Accept })
```
skazi0: it should be `==` not `=`.

skazi0 (changes requested):

assets/app/data/crowbar/services/crowbar-backup.factory.spec.js
MaximilianMeister: fixed

skazi0 (approved):

assets/app/data/crowbar/services/crowbar-backup.factory.js
santiph: +1

assets/app/data/crowbar/services/crowbar-backup.factory.spec.js
```diff
@@ -36,7 +36,8 @@ describe('Crowbar Backup Factory', function () {
         describe('when create method is executed', function () {
 
             beforeEach(function () {
-                $httpBackend.expect('POST', '/utils/backups', undefined)
+                $httpBackend.expect('POST', '/api/upgrade/adminbackup', undefined, function (headers) {
+                    return headers.Accept == COMMON_API_V2_HEADERS.Accept })
```
santiph: Please, use identity comparison here.
http://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons

santiph (changes requested):
Please, fix the comments

assets/app/data/crowbar/services/crowbar-backup.factory.js
skazi0: Actually I withdrew this comment. 
Naming services by URL component doesn't seem to work very well for us.
Changes in API structure lead to unnecessary code changes on UI side which can introduce additional errors.
`crowbarBackupFactory` is a great example: it uses `/utils/backups` and `/api/upgrade/adminbackup` which with our current/old approach would cause these functions to be located in two services. IMO it's better to have them grouped by the functional area they cover (i.e. backups in this case).

santiph: The taken approach, with exception of backup, is to reflect the entities exposed by the API and let each controller decide which ones needs to be consumed during the upgrade flow.
The problem here are not the changes we need to do in the UI, but the uncompleted design of the API that force us as client to re-work our backend integration layer.

santiph (approved):

skazi0 (changes requested):
I don't think this 1:1 mapping of backend to services is a good approach. The services should create abstractions on top of API which are meaningful from UI point of view.

Backups are great example of this. They are handled in two places in the backend but from UI side they are one thing and service/factory should take care of creating this concept for the controllers.

assets/app/data/crowbar/services/crowbar-backup.factory.js
Itxaka: +1 to @santiph approach . This seems to stem from an ongoing development of the API parallel to the UI instead of working with an stable/backwards compatible API. `¯\_(ツ)_/¯`

MaximilianMeister:
can we then please create a followup for the commit in discussion? I can't easily update changes from @santiph here. so I would propose to remove the commit from here and @santiph creates a follow up PR

@santiph would that be ok for you?

Itxaka (approved):

skazi0:
OK. 2/3 votes for @santiph so I have no choice. I still don't like this approach though.

skazi0 (approved):

PR #88, opened 2016-11-24
upgrade: increase coverage of upgrade controller

Itxaka:
- create 2 different describe sections for nextstep and isLastStep
 - nextStep tests
   - when called it should move to the next step
   - when called in the last step it should not change steps
 - isLastStep tests
   - when called on the first step it should return false
   - when called on the last step it should return true

Before:
```
Statements   : 99.06% ( 1482/1496 )
Branches     : 86.36% ( 57/66 )
Functions    : 98.5% ( 590/599 )
Lines        : 99.06% ( 1482/1496 )
```

After:
```
Statements   : 99.34% ( 1508/1518 )
Branches     : 89.39% ( 59/66 )
Functions    : 98.84% ( 598/605 )
Lines        : 99.34% ( 1508/1518 )
```

### DIFF
--- a/assets/app/features/upgrade/upgrade.controller.spec.js
+++ b/assets/app/features/upgrade/upgrade.controller.spec.js
@@ -1,4 +1,4 @@
-/*global bard $controller $httpBackend should assert upgradeStepsFactory $rootScope*/
+/*global bard $controller $httpBackend should assert upgradeStepsFactory $rootScope $state*/
 describe('Upgrade Controller', function () {
     var controller,
         controllerScope;
@@ -13,8 +13,18 @@ describe('Upgrade Controller', function () {
 
         controllerScope = $rootScope.$new();
 
+        //Mock the $state service
+        bard.mockService($state, {
+            go : true
+        });
+
+        //Spy on the go function
+        spyOn($state, 'go');
+
+        $state.current = {id: 0, name: 'upgrade.backup', state: 'upgrade.backup'};
+
         //spy on the refeshStepsList function
-        spyOn(upgradeStepsFactory, 'refeshStepsList');
+        spyOn(upgradeStepsFactory, 'refeshStepsList').and.callThrough();
 
         //Create the controller
         controller = $controller('UpgradeController', {'$scope': controllerScope});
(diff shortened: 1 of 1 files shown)


skazi0 (approved):
lgtm

assets/app/features/upgrade/upgrade.controller.spec.js
```diff
@@ -44,14 +54,51 @@ describe('Upgrade Controller', function () {
...
-            expect(controller.steps.isLastStep).toEqual(jasmine.any(Function));
+        describe('isLastStep function', function () {
+            it('should exist', function () {
+                should.exist(controller.steps.isLastStep);
+                expect(controller.steps.isLastStep).toEqual(jasmine.any(Function));
+            });
+
+            it('when called on the first step it should return false', function () {
+                // set the first step in the list as current
+                $state.current = {id: 0, name: 'upgrade.backup', state: 'upgrade.backup'};
```
santiph, line 91: please, use UPGRADE_STEPS constant

santiph (changes requested):

assets/app/features/upgrade/upgrade.controller.spec.js
```diff
@@ -44,14 +54,51 @@ describe('Upgrade Controller', function () {
...
+                var nextStep = (controller.steps.list[upgradeStepsFactory.activeStep.id + 1]).state;
+                controller.steps.nextStep();
+                expect($state.go).toHaveBeenCalledWith(nextStep);
+            });
+
+            it('when called in the last step it should not change steps', function () {
+                // set the last step in the list as current
+                $state.current = {id: 0, name: 'upgrade.upgrade-nodes', state: 'upgrade.upgrade-nodes'};
+                // refresh the list so it sets it as active
+                upgradeStepsFactory.refeshStepsList();
```
santiph: `upgradeStepsFactory` is an external component that should be tested independently. In order to unit test the upgradeController, this factory needs to be mocked/injected and manipulated as part of the preconditions.

So instead of `upgradeStepsFactory.refeshStepsList();`, I think the following could be done:
```
upgradeStepsFactory.activeStep = upgradeStepsFactory.steps.list[upgradeStepsFactory.steps.list.length - 1];
```

santiph: I'd even consider going one step further and mock the response for `isLastStep()` function to return true/false according to the scenario being tested, and ignore completely how the last step is being determined.

assets/app/features/upgrade/upgrade.controller.spec.js
```diff
@@ -44,14 +54,51 @@ describe('Upgrade Controller', function () {
...
+                // set the first step in the list as current
+                $state.current = {id: 0, name: 'upgrade.backup', state: 'upgrade.backup'};
+                expect(controller.steps.isLastStep()).toBe(false);
+            });
+
+            it('when called on the last step it should return true', function () {
+                // set the last step in the list as current
+                $state.current = {id: 0, name: 'upgrade.upgrade-nodes', state: 'upgrade.upgrade-nodes'};
+                // refresh the list so it sets it as active
+                upgradeStepsFactory.refeshStepsList();
```
santiph: As mentioned above, I'd replace this with 
`upgradeStepsFactory.activeStep = upgradeStepsFactory.steps.list[upgradeStepsFactory.steps.list.length - 1];`

Attribute `id` should also be updated

santiph (changes requested):
Good improvement overall. Please review the comments

assets/app/features/upgrade/upgrade.controller.spec.js
Itxaka: But mocking that part means that if it ever changes, the tests will continue to run for `nextStep` and `isLastStep` and they will return as valid. Both functions require that `upgradeStepsFactory.refeshStepsList()` works as intended to work properly so not sure why we will mock a dependent service instead of relaying on the real one to be sure that it works?

Mocking is the devil, it can hide real issues if you mock too much so I try to mock only the necessary stuff, hence the change to ```spyOn(upgradeStepsFactory, 'refeshStepsList').and.callThrough();``` as we need the *real* service to work for the other functions to do so.

assets/app/features/upgrade/upgrade.controller.spec.js
Itxaka: Not sure I understand this. `UPGRADE_STEPS` do not contains the state names that we need to test this properly no?

assets/app/features/upgrade/upgrade.controller.spec.js
santiph: Such problem should be validated on the service unit test, not this. Is not a responsibility of this controller to refresh the list of steps.
We need to enforce as many test scenarios as possible to make our test to fail if something goes wrong, but just to ensure each function does what is supposed to.

I think the way things are integrated and supposed to be working all together will be covered by automation tests.

assets/app/features/upgrade/upgrade.controller.spec.js
santiph: You're right. I was confused with the steps names returned from the service.

santiph (approved):

PR #90, opened 2016-11-28
enable and document docker development environment

MaximilianMeister:
this makes it possible to run a simple and reusable devsetup in a
contained environment that includes all dependencies

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM google/nodejs
+MAINTAINER Maximilian Meister <mmeister@suse.de>
+
+ENV NODE_ENV=development
+ENV PORT=3000
+
+WORKDIR /crowbar-ui
+
+RUN apt-get update && \
+    apt-get install -y libfontconfig1 libfontconfig1-dev supervisor
+RUN npm install --global gulp bower nodemon
+RUN mkdir -p /var/log/supervisor
+COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
+
+ADD . .
+
+RUN npm install && \
+    bower --allow-root install && \
+    npm build
+
+EXPOSE 3000
+
+ENTRYPOINT ["/usr/bin/supervisord"]
--- a/README.md
+++ b/README.md
@@ -1,7 +1,30 @@
 # Welcome to Crowbar Angular Prototype
 [![Build Status](https://travis-ci.org/crowbar/crowbar-ui.svg?branch=master)](https://travis-ci.org/crowbar/crowbar-ui)
 
-##### First steps to have the app running locally for PRODUCTION ENV only:
+### Setting up a development environment with Docker
+
+This is an optional way to develop on `crowbar-ui` within a contained environment. The main benefit is to have some reusable and host-independent environment to develop on.
+
+1. make sure `docker` is installed and the `docker` daemon is running.
+
+   for installation details see: https://docs.docker.com/engine/installation/
+2. build the image from the [Dockerfile](Dockerfile):
+  ```
+  docker build -t crowbar-ui .
+  ```
+3. run the container (`pwd` should be your git clone):
+  ```
+  docker run -d --name "crowbar-ui_container" -v $(pwd)/assets:/crowbar-ui/assets -v $(pwd)/routes:/crowbar-ui/routes -p 3000:3000 crowbar-ui
+  ```
+4. work on the assets folder in your local git clone and the changes will be immediately visible in the browser.
+   also if you are changing the `./routes` directory the server inside the container gets restarted by `nodemon`
+
+#### Running the test suite inside the container
+```
+docker exec -ti crowbar-ui_container npm test
+```
+
+### First steps to have the app running locally for PRODUCTION ENV only:
 
 1 - Clone the app from GitHub:
 
--- a/supervisord.conf
+++ b/supervisord.conf
@@ -1,0 +1,14 @@
+[supervisord]
+nodaemon=true
+
+[program:gulp]
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
+autorestart=true
+command=/bin/bash -c "/crowbar-ui/node_modules/gulp/bin/gulp.js"
+
+[program:nodemon]
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
+autorestart=true
+command=/bin/bash -c "nodemon /crowbar-ui/bin/www.js"
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+Dockerfile
+.git


Itxaka:
Looks good. The only issue I can find its that as long as `npm start` is running, if you change something in the routes, it wont be reflected unless `npm start` is relaunched (happens to me locally) so any changes to the routes require relaunching the container, but thats a minor issue.

Also, would be good to have gulp already running on container start instead of manually launching it.

MaximilianMeister:
> Looks good. The only issue I can find its that as long as npm start is running, if you change something in the routes, it wont be reflected unless npm start is relaunched (happens to me locally) so any changes to the routes require relaunching the container, but thats a minor issue.

so the routes are the mocked back-end right? then maybe I should rephrase a bit?
```
Setting up a UI development environment with Docker
```

> Also, would be good to have gulp already running on container start instead of manually launching it.

good idea, i thought about that too, but apparently as there can only be one `ENTRYPOINT` and the two tasks are both foreground tasks it may require another linked container to be really clean. but one could also hack some bash line/script that runs both tasks in one entrypoint, but i find it a bit dirty.

Itxaka:
> so the routes are the mocked back-end right? then maybe I should rephrase a bit?

Yup! Just something to have in mind when testing and maybe changing the fake backend.


> good idea, i thought about that too, but apparently as there can only be one ENTRYPOINT and the two tasks are both foreground tasks it may require another linked container to be really clean. but one could also hack some bash line/script that runs both tasks in one entrypoint, but i find it a bit dirty.

Yeah. I would recommend using [supervisord](https://docs.docker.com/engine/admin/using_supervisord/) or something like that in order to have multiple services running in a single container. Unfortunately it requires an extra file for config, but its the best solution I have find for this.

Itxaka (approved):
My complains are just nitpicks, not blocking.

README.md
```diff
@@ -1,6 +1,23 @@
 # Welcome to Crowbar Angular Prototype
 [![Build Status](https://travis-ci.org/crowbar/crowbar-ui.svg?branch=master)](https://travis-ci.org/crowbar/crowbar-ui)
 
+### Setting up a development environment with Docker
+
+1. make sure `docker` is installed and the `docker` daemon is running.
```
santiph, line 8: Instructions for this

README.md
```diff
@@ -1,6 +1,23 @@
...
+```
+3. run the container:
+```
+docker run -d --name "crowbar-ui_container" -v $(pwd)/assets:/crowbar-ui/assets -p 3000:3000 crowbar-ui
+```
+4. run the `gulp` watchdog:
+```
+docker exec -ti crowbar-ui_container gulp
+```
+5. work on the assets folder in your local git clone and the changes will be immediately visible in the browser.
```
santiph: assets is only one of the folders required for this project. If the container will run the web server, then we need to also synchronise routes folder and app.js file to touch the mocked endpoints and their responses.
Also files in the root level provides the configuration for development and infrastructure related work. package.json and bower.json files are crucial to include new dependencies. How will them be synchronised?

Also, how are unit tests going to be run here, if the dependencies are installed in the container, not the host? And we need access to the coverage folder in order to get the coverage reports.

santiph (dismissed):
I think the requirements for this container are not very clear.
I don't really see the benefit. Either way we need to install a dependency in the host machine, either docker or nodejs

README.md
MaximilianMeister: why should i rewrite the docs in this repo on how to set up docker? i think anyone who is interested in using docker for the dev env should not need this info.

README.md
MaximilianMeister: i didn't know that, probably due to the missing/outdated documentation on how to setup the dev env with all dependencies on the host.

can you please document what exactly is to be done when you change other files than the assets folder?

regarding the unit tests, can't we just run them inside the container? we have all necessary files there i believe.

MaximilianMeister:
> I don't really see the benefit. Either way we need to install a dependency in the host machine, either docker or nodejs

you don't need to use docker, there is still the option to set it up on the host itself, but documentation needs to be updated ideally from someone who knows all the details

README.md
Itxaka: re: unit tests

They can be run on teh container and if we use supervisord for them, we can make them run on every change with the `--watch` option. And if we mount the current dir instead of the assets folder, we can have the coverage report available on the host :D

MaximilianMeister: yes i thought about adding the whole directory too, but then it would be a bit useless to have a container, because everything will be mirrored :-( and the coverage folder is generated, so that shouldn't be a problem I guess.

now that we have a third usecase I might try out `supervisord`

MaximilianMeister: btw can't we just make a `docker exec -ti crowbar-ui_container npm test` and get the result?

Itxaka: > btw can't we just make a docker exec -ti crowbar-ui_container npm test and get the result?

Yup! The only problem is that the coverage files are generated on the container so its not easy to see them locally in the browser. (karma generates a full coverage report in html very useful to see what files/lines/branches you are missing in the different dirs/files)

> then it would be a bit useless to have a container, because everything will be mirrored 

Umm, yeah. It depends on what is the objective of using docker here. Is it to have the software installed independently from the host? Is it to run it isolated?

MaximilianMeister: i see. how do you usually access the coverage report then? is it served via express.js as well?

you're probably right about the objective, it is not really clear yet. it should just be as easy as possible to quickly start developing, throwing away, and rebuilding ;-) and a container can be great for that. 

but anyways I guess this is still much WIP, so I'll try around a bit (currently with your suggestion `supervisord`)

Itxaka: > i see. how do you usually access the coverage report then? is it served via express.js as well?

Nah, its just plain html files, nothing extra needed

Itxaka (changes requested):
Wip!

README.md
MaximilianMeister: hmm, if it would be served too, we could just access it at some url from within the container

Itxaka: can probably serve it with `python -m SimpleHTTPServer 8000` trougth supervisord :grimacing: 

Should be able to serve it through nodejs but I dont know if you can do so without an app.js

README.md
santiph: From my point of view, a development setup guide should guide the user to set up the environment as fast and easily as possible.
Two or three steps detailing how to install docker and setup the daemon (if required) would help any user of this guide to get the thing up and running

README.md
santiph: An extra express server should be configured to do so, but there is no need to be server through a web sever. It can be opened with the regular browser

MaximilianMeister: the problem here is that the testing creates a subdirectory under `/coverage/PhantomJS\ 2.1.1\ \(Linux\ 0.0.0\)`. it can be served by `express` if the folder is registered as a static folder. but how can i know this value? it seems to be dynamic (with the version of phantomjs)

it should be something like:
```js
if (app.get('env') === 'development') {
    app.use(express.static(__dirname + '/coverage/' + get_name_of_that_directory()));
}
```

then it could easily be accessed via `localhost:3000/coverage`

santiph: Coverage reporter uses the name and version of the browser used to name the subfolder.
I have `PhantomJS 2.1.1 (Mac OS X 0.0.0)` and `Chrome 54.0.2840 (Mac OS X 10.12.0)` subfolders in /coverage

santiph: Also, for debugging purposes, this command is used: `./node_modules/karma/bin/karma start --browsers Chrome`, which opens an isolated chrome instance where to run all the unit tests and debug them  through the development tools.
Is it possible to access this within the container?

MaximilianMeister: no, i guess in the container we can't open chrome or any browser/webkit other than PhantomJS. it needs to be headless

Itxaka: You can run xvfb to run chrome in a headless environment, though you miss the ability to see the browser and debug, you still get to run the tests in different browsers :)

Dockerfile
```diff
@@ -6,6 +6,10 @@ ENV PORT=3000
 
 WORKDIR /crowbar-ui
 VOLUME ./assets
+RUN apt-get update && \
+    apt-get install -y libfontconfig1 libfontconfig1-dev openssh-server apache2 supervisor
```
Itxaka: ssh and apache? leftover from testing?

MaximilianMeister: this is from the official docker docs of `supervisord` (https://docs.docker.com/engine/admin/using_supervisord/).

not sure either why it has to be there, maybe supervisord uses those somehow

Itxaka: Ah no, that because on their example they are running both ssh and apache trougth supervisord, so they have to install them. We should remove them from the dockerfile as they are not needed for supervisord to run.

MaximilianMeister: ok, thanks for the closer look there ;)

MaximilianMeister:
I have used `nodemon` now to restart the server once some file in `routes` changed. please have another look, especially if i need to monitor other folders too for restarting the server.
i know if you add new dependencies it might not work, but for that case you have to rebuild the container image i'd say.

it still would be nice to serve the coverage through some accessible url or port. maybe I'll continue trying that. but this PR is in a usable state now IMO, follow ups can always be done later.

README.md
```diff
@@ -1,6 +1,24 @@
...
+1. make sure `docker` is installed and the `docker` daemon is running.
+
+   for installation details see: https://docs.docker.com/engine/installation/
+2. build the image from the [Dockerfile](Dockerfile):
+  ```
+  docker build -t crowbar-ui .
+  ```
+3. run the container (`pwd` should be your git clone):
+  ```
+  docker run -d --name "crowbar-ui_container" -v $(pwd)/assets:/crowbar-ui/assets -v $(pwd)/routes:/crowbar-ui/routes -p 3000:3000 -p 3001:9876 crowbar-ui
```
Itxaka: Not sure why there is a second set of ports linked here?

supervisord.conf
```diff
@@ -0,0 +1,8 @@
+[supervisord]
+nodaemon=true
+
+[program:gulp]
+command=/bin/bash -c "/crowbar-ui/node_modules/gulp/bin/gulp.js"
```
Itxaka, line 8: Can I suggest the use of:

```
stdout_logfile=/dev/stdout
stdout_logfile_maxbytes=0
```

on the program(s) section so we can see the output of all programs with docker logs $CONTAINER_ID?

Also, using `autorestart=true` may be good in case we break something and fix it so we dont need to restart the container and its restarted automagically :)

Itxaka (changes requested):
Tested locally, works as expected. A couple of improvements missing.

README.md
MaximilianMeister: oops, yes, that still is a remainder of trying to get the coverage static files served... and karma seems to have some open port in the config, and i thought it is a webserver for the coverage, but it didn't work somehow...

supervisord.conf
MaximilianMeister: ok, i think autorestart is default (but i need to check). at least when i killed some process inside, it respawned automatically

Itxaka (approved):
Looking good to me!

MaximilianMeister:
@santiph do you still have a change request here?

sidthe:
@santiph can this PR be closed?

santiph:
After discussing with @kwwii, this proposal will be reevaluated after the next release.

MaximilianMeister:
> After discussing with @kwwii, this proposal will be reevaluated after the next release.

What is there to evaluate for a release? It's a documentation about developing on `crowbar-ui` with docker, which doesn't get shipped within the release

rsalevsky (approved):

santiph:
Because adding these changes now would also effect the local testing process we are currently using as well as the fact that this itself is not tested by CI, I'd like to reevaluate this proposal properly to come up with an unified development environment after cloud 7 is released.

santiph:
Until a unified solution is defined, would you consider creating a separate to create a docker specific repository and use `crowbar-ui` as a dependency?

rsalevsky:
@santiph In the OpenSource world It's nearly impossible to have a unified deployment way. If we would be strict here then we would have to use the packages from the build service and everything else would not count. So having more possible options how to start developing is in my opinion create due to there will be never the optimal solution which fits everyone's needs.

mjura (approved):

mjura (approved):
+1

skazi0:
> Because adding these changes now would also effect the local testing process we are currently using (...)

I don't see how these changes affect our current way of testing the app. It's up to the developer if he/she wants to use docker or traditional setup.

Itxaka:
@skazi0 @jsuchome should we update or  close this already?

skazi0:
@Itxaka as this is only documentation, I think it's safe to merge it.

rsalevsky:
+1 for merging this

PR #86, opened 2016-11-22
upgrade: Updated URL for node repo checks

skazi0:

PR #91, opened 2016-11-28
upgrade: Fix dummy backend URL for noderepochecks

skazi0:
Follow up on https://github.com/crowbar/crowbar-ui/pull/86 wrt the dummy backend used for development.


Itxaka:
shouldnt we rename the file and change line 13 as well?

skazi0:
The consistency is not maintained in this part. Check `/api/upgrade/adminrepocheck` above. This code doesn't go into the product so I see no point in spending much too much effort on this.

Itxaka (approved):
Thanks!

santiph:
I'd still ask to keep consistency and rename the files.
Even when it's not going to the final product, is supposed to ease our dev work and therefore I think it's worth to maintain it. Besides, it should be a quick file rename, not huge refactor

PR #93, opened 2016-11-28
upgrade: increase coverage for upgrade-status factory

Itxaka:
Checks for the path when the response is rejected

### DIFF
--- a/assets/app/data/crowbar/services/upgrade-status.factory.spec.js
+++ b/assets/app/data/crowbar/services/upgrade-status.factory.spec.js
@@ -187,6 +187,61 @@ describe('Upgrade Status Factory', function () {
                     });
                 });
             });
+
+            describe('when got upgrade status from api unsuccessfully', function () {
+                describe('when timeout == 0', function () {
+                    beforeEach(function () {
+                        bard.mockService(upgradeFactory, {
+                            getStatus: $q.reject()
+                        });
+
+                        upgradeStatusFactory.waitForStepToEnd(
+                            testedStep, mockedSuccessCallback, mockedErrorCallback, pollingInterval
+                        );
+
+                        $rootScope.$digest();
+                    });
+
+                    it('should not call success callback', function () {
+                        expect(mockedSuccessCallback).not.toHaveBeenCalled();
+                    });
+
+                    it('should call error callback', function () {
+                        expect(mockedErrorCallback).toHaveBeenCalled();
+                    });
+
+                    it('should not schedule another check', function () {
+                        expect(mockedTimeout).not.toHaveBeenCalled();
+                    });
+                });
+
+                describe('when timeout > 0', function () {
+                    beforeEach(function () {
+                        bard.mockService(upgradeFactory, {
+                            getStatus: $q.reject()
+                        });
+
+                        upgradeStatusFactory.waitForStepToEnd(
+                            testedStep, mockedSuccessCallback, mockedErrorCallback, 1, 10
+                        );
+
+                        $rootScope.$digest();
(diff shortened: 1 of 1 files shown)


skazi0 (approved):

MaximilianMeister (approved):
looks good as far as i can judge, and tests passed

assets/app/data/crowbar/services/upgrade-status.factory.spec.js
```diff
@@ -187,6 +187,63 @@ describe('Upgrade Status Factory', function () {
...
+                    beforeEach(function () {
+                        bard.mockService(upgradeFactory, {
+                            getStatus: $q.reject()
+                        });
+
+                        upgradeStatusFactory.waitForStepToEnd(
+                            testedStep, mockedSuccessCallback, mockedErrorCallback, 1, 10
+                        );
+
+                        spyOn(upgradeStatusFactory, 'waitForStepToEnd').and.callThrough();
```
santiph: Why are you spying on this method?
Shouldn't we validate it has been called?

Itxaka: Ah yes, good point! 

Its a leftover. I was trying to validate to see if it was being called, but I was getting that it wasnt called. 

I migth need to call $scope.$digest() or something like that due to this function being called by setTimeout?

Im not sure how to proceed to check if that was called tbh.

assets/app/data/crowbar/services/upgrade-status.factory.spec.js
```diff
@@ -187,6 +187,61 @@ describe('Upgrade Status Factory', function () {
...
+
+                        $rootScope.$digest();
+                    });
+
+                    it('should not call success callback', function () {
+                        expect(mockedSuccessCallback).not.toHaveBeenCalled();
+                    });
+
+                    it('should call error callback', function () {
+                        expect(mockedErrorCallback).toHaveBeenCalled();
```
santiph: Please, validate it has been called once. Also, it would be great to also assert the right `errorResponse` is being sent as a parameter
https://github.com/Itxaka/crowbar-ui/blob/c6682481862e7d0c2bb2e4ff3ce9b3837b2ab25a/assets/app/data/crowbar/services/upgrade-status.factory.js#L49

assets/app/data/crowbar/services/upgrade-status.factory.spec.js
```diff
@@ -187,6 +187,61 @@ describe('Upgrade Status Factory', function () {
...
+                    it('should not call success callback', function () {
+                        expect(mockedSuccessCallback).not.toHaveBeenCalled();
+                    });
+
+                    it('should not call error callback', function () {
+                        expect(mockedErrorCallback).not.toHaveBeenCalled();
+                    });
+
+                    it('should schedule another check', function () {
+                        expect(mockedTimeout).toHaveBeenCalled();
```
santiph: Is it possible to assert the number of times the function is being called?
Also, the two parameters that are being sent.

santiph (changes requested):
Please, check if the assertions could be improved

assets/app/data/crowbar/services/upgrade-status.factory.spec.js
Itxaka: done

assets/app/data/crowbar/services/upgrade-status.factory.spec.js
Itxaka: done

santiph (approved):

PR #95, opened 2016-11-29
upgrade: Landing page refresh handling

skazi0:
When "prepare" step is in progress UI polls for status to detect
when step is finished. The polling is started when user clicks
"Begin Upgrade" button.

When page is refreshed or opened while the step is already running
the UI needs to automatically switch to "waiting" state.

### DIFF
--- a/assets/app/data/crowbar/services/upgrade-status.factory.spec.js
+++ b/assets/app/data/crowbar/services/upgrade-status.factory.spec.js
@@ -161,11 +161,11 @@ describe('Upgrade Status Factory', function () {
                 });
 
                 it('should call completed callback', function () {
-                    expect(mockedCompletedCallback).toHaveBeenCalled();
+                    expect(mockedCompletedCallback).toHaveBeenCalledTimes(1);
                 });
 
                 it('should not call running callback', function () {
-                    expect(mockedRunningCallback).not.toHaveBeenCalled();
+                    expect(mockedRunningCallback).not.toHaveBeenCalledTimes(1);
                 });
 
                 it('should set completed flag to true', function () {
--- a/assets/app/data/crowbar/services/upgrade-status.factory.spec.js
+++ b/assets/app/data/crowbar/services/upgrade-status.factory.spec.js
@@ -1,4 +1,4 @@
-/*global bard $q $rootScope should expect module upgradeFactory upgradeStatusFactory */
+/*global bard assert $q $rootScope should expect module upgradeFactory upgradeStatusFactory */
 describe('Upgrade Status Factory', function () {
     var pollingInterval = 1234,
         testedStep = 'admin_upgrade',
@@ -102,8 +102,11 @@ describe('Upgrade Status Factory', function () {
         incompleteUpgradeResponse = {
             data: incompleteUpgradeResponseData
         },
+        flagsObject,
         mockedSuccessCallback,
         mockedErrorCallback,
+        mockedRunningCallback,
+        mockedCompletedCallback,
         mockedTimeout,
         errorResponse = {
             error: 'some error'
@@ -115,6 +118,8 @@ describe('Upgrade Status Factory', function () {
 
         mockedSuccessCallback = jasmine.createSpy('onSuccess');
         mockedErrorCallback = jasmine.createSpy('onError');
+        mockedCompletedCallback = jasmine.createSpy('onCompleted');
+        mockedRunningCallback = jasmine.createSpy('onRunning');
 
         // mock $timeout service
         module(function($provide) {
@@ -132,6 +137,77 @@ describe('Upgrade Status Factory', function () {
 
         it('returns an object', function () {
             should.exist(upgradeStatusFactory);
+        });
+
+        describe('syncStatusFlags function', function () {
+            it('should be defined', function () {
+                should.exist(upgradeStatusFactory.syncStatusFlags);
+                expect(upgradeStatusFactory.syncStatusFlags).toEqual(jasmine.any(Function));
+            });
+
+            describe('when received status is completed', function () {
+                beforeEach(function () {
+                    bard.mockService(upgradeFactory, {
+                        getStatus: $q.when(completedUpgradeResponse)
+                    });
+
+                    flagsObject = { running: true, completed: false };
+
+                    upgradeStatusFactory.syncStatusFlags(
+                        testedStep, flagsObject, mockedRunningCallback, mockedCompletedCallback
+                    );
+
+                    $rootScope.$digest();
+                });
+
+                it('should call completed callback', function () {
+                    expect(mockedCompletedCallback).toHaveBeenCalled();
+                });
+
+                it('should not call running callback', function () {
+                    expect(mockedRunningCallback).not.toHaveBeenCalled();
+                });
+
+                it('should set completed flag to true', function () {
+                    assert.isTrue(flagsObject.completed);
+                });
+
+                it('should set running flag to false', function () {
+                    assert.isFalse(flagsObject.running);
+                });
+            });
+
+            describe('when received status is running', function () {
+                beforeEach(function () {
+                    bard.mockService(upgradeFactory, {
+                        getStatus: $q.when(incompleteUpgradeResponse)
+                    });
+
+                    flagsObject = { running: false, completed: true };
+
+                    upgradeStatusFactory.syncStatusFlags(
+                        testedStep, flagsObject, mockedRunningCallback, mockedCompletedCallback
+                    );
+
+                    $rootScope.$digest();
+                });
+
+                it('should not call completed callback', function () {
+                    expect(mockedCompletedCallback).not.toHaveBeenCalled();
+                });
+
+                it('should call running callback', function () {
+                    expect(mockedRunningCallback).toHaveBeenCalled();
+                });
+
+                it('should set completed flag to false', function () {
+                    assert.isFalse(flagsObject.completed);
+                });
+
+                it('should set running flag to true', function () {
+                    assert.isTrue(flagsObject.running);
+                });
+            });
         });
 
         describe('waitForStepToEnd function', function () {
--- a/assets/app/data/crowbar/services/upgrade-status.factory.spec.js
+++ b/assets/app/data/crowbar/services/upgrade-status.factory.spec.js
@@ -3,7 +3,7 @@ describe('Upgrade Status Factory', function () {
     var pollingInterval = 1234,
         testedStep = 'admin_upgrade',
         completedUpgradeResponseData = {
-            current_step: 'admin_upgrade',
+            current_step: 'database',
             substep: null,
             current_node: null,
             steps: {
@@ -165,7 +165,7 @@ describe('Upgrade Status Factory', function () {
                 });
 
                 it('should not call running callback', function () {
-                    expect(mockedRunningCallback).not.toHaveBeenCalledTimes(1);
+                    expect(mockedRunningCallback).not.toHaveBeenCalled();
                 });
 
                 it('should set completed flag to true', function () {
@@ -197,7 +197,7 @@ describe('Upgrade Status Factory', function () {
                 });
 
                 it('should call running callback', function () {
-                    expect(mockedRunningCallback).toHaveBeenCalled();
+                    expect(mockedRunningCallback).toHaveBeenCalledTimes(1);
                 });
 
                 it('should set completed flag to false', function () {
--- a/assets/app/features/upgrade/controllers/upgrade-landing.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-landing.controller.js
@@ -20,7 +20,6 @@
         'ADDONS_PRECHECK_MAP',
         'PREPARE_TIMEOUT_INTERVAL',
         'UPGRADE_STEPS',
-        'STEP_STATES'
     ];
     // @ngInject
     function UpgradeLandingController(
@@ -31,8 +30,7 @@
         crowbarFactory,
         ADDONS_PRECHECK_MAP,
         PREPARE_TIMEOUT_INTERVAL,
-        UPGRADE_STEPS,
-        STEP_STATES
+        UPGRADE_STEPS
     ) {
         var vm = this,
             optionalPrechecks = {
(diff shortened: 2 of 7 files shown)


skazi0:
@Itxaka, @santiph looks like I finally found a good way for selective re-mocking of services for testing special cases :)

skazi0:

assets/app/features/upgrade/controllers/upgrade-landing.controller.js
```diff
@@ -86,6 +88,24 @@
...
+            // adjust UI state to backend status
+            upgradeFactory.getStatus()
+                .then(
+                    function (response) {
+                        // skz: This part is disabled on purpose. There's no easy way to restore complete checks state
+                        // without running the checks again so it's better to leave this to the user.
+                        //vm.prechecks.running = response.data.steps.upgrade_prechecks.status === STEP_STATES.running;
+                        //vm.prechecks.completed = response.data.steps.upgrade_prechecks.status === STEP_STATES.passed;
+
+                        vm.prepare.running = response.data.steps.upgrade_prepare.status === STEP_STATES.running;
```
santiph: Could this method be stored in a helper?

assets/app/features/upgrade/controllers/upgrade-landing.controller.js
```diff
@@ -121,6 +129,22 @@
             );
         }
 
+        function waitForPrepareToEnd() {
```
santiph, line 119: Missing doc block

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
```diff
@@ -79,6 +211,69 @@ describe('Upgrade Landing Controller', function() {
...
+        it('prepare should not be completed', function() {
+            assert.isFalse(controller.prepare.completed);
+        });
+    });
+
+    describe('when created while prepare is running', function() {
+        beforeEach(function() {
+            // local change in mocked service
+            upgradeFactory.getStatus =
+                sinon.spy(function () { return $q.when(incompleteUpgradeResponse); });
```
santiph: Why have you used sinon and not jasmine's spy to return values?
`spyOn(upgradeFactory, 'getStatus').and.returnValue($q.when(incompleteUpgradeResponse));`

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
```diff
@@ -79,6 +211,69 @@ describe('Upgrade Landing Controller', function() {
...
+
+            controller = $controller('UpgradeLandingController');
+
+            $rootScope.$digest();
+        });
+
+        it('should start polling for status', function() {
+            expect(upgradeStatusFactory.waitForStepToEnd).toHaveBeenCalled();
```
santiph: Is it possible to validate the number of time it has been called?

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
```diff
@@ -79,6 +211,69 @@ describe('Upgrade Landing Controller', function() {
         should.exist(controller);
     });
 
+    describe('when created while prepare is not running', function() {
```
santiph, line 214: "Upgrade Landing Controller when created..." sounds a bit weird... Also, shouldn't this be "before running prepare nodes" instead of "while it's not running"?

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
```diff
@@ -79,6 +211,69 @@ describe('Upgrade Landing Controller', function() {
...
+        it('prepare should be running', function() {
+            assert.isTrue(controller.prepare.running);
+        });
+
+        it('prepare should not be completed', function() {
+            assert.isFalse(controller.prepare.completed);
+        });
+    });
+
+    describe('when created while prepare is finished', function() {
```
santiph: "after prepare nodes finished"?

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
```diff
@@ -79,6 +211,69 @@ describe('Upgrade Landing Controller', function() {
...
+        beforeEach(function() {
+            // local change in mocked service
+            upgradeFactory.getStatus =
+                sinon.spy(function () { return $q.when(completedUpgradeResponse); });
```
santiph: same question regarding jasmine's spy

santiph (commented):
Minor comments. Please consider using jasmine's spy to keep consistency with other unit tests, unless there is any other limitation to switch to sinon

assets/app/features/upgrade/controllers/upgrade-landing.controller.js
skazi0: It's just 4 lines of code.

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
skazi0: No. "before running prepare nodes" doesn't express what is being tested here. The key is that we want to check the controller creation/activation (i.e. page load) in different cases. The "describe" parts are destined to sound weird sooner or later when you start nesting things.

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
skazi0: Don't know... spying on a spy seems a bit redundant. Are we playing double agents here?

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
skazi0: I don't think this is important in this context but I can add it.

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
skazi0: Changed it to "when created after prepare is finished". This is to express that we're testing the creation procedure not the behavior of controller when the prepare step is finished.

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
skazi0: same answer.

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
santiph: Haha nono, I was thinking on overriding the original spy. 😛 
Actually, I'm not sure if my first snippet works. I haven't tried yet.
But an alternative to keep using jasmine could be:
`upgradeFactory.getStatus = jasmine.createSpy('getStatus').and.returnValue($q.when(incompleteUpgradeResponse))`

skazi0: My implementation is direct extract of what bardJS is doing inside `mockService`. Using jasmine here would cause the service mock to have inconsistent implementation. Some functions would be mocked using sinon and some using jasmine. Not sure if the interfaces are 100% complatible. If not, then using those spies in expectations would require different syntax depending on the library used for particular function.

Anyway, I was thinking about a different approach. Why don't we create an utility library for our tests (e.g. SUSE Test Utils = stu)? The call would then be something like: `stu.remockService(upgradeFactory, { getStatus: $q.when(incompleteUpgradeResponse) } )`. This would mimic the `mockService()` syntax from bardJS and could be reused in other places.

assets/app/features/upgrade/controllers/upgrade-landing.controller.js
santiph: It is, but checking if a status of each step is something we would be using across the entire upgrade flow, not just in the landing controller...

skazi0: I thought you mean just extracting this code into a local function. If you want to make it somehow reusable then it  makes more sense. I'll think about it.

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
```diff
@@ -1,8 +1,137 @@
...
-  crowbarFactory $q $rootScope module $state */
+  crowbarFactory $q $rootScope module $state sinon */
 describe('Upgrade Landing Controller', function() {
     var controller,
+        completedUpgradeResponseData = {
+            current_step: 'upgrade_prepare',
+            substep: null,
+            current_node: null,
+            steps: {
+                upgrade_prechecks: {
```
santiph, line 11: Same comment as https://github.com/crowbar/crowbar-ui/pull/97/files#r90652293

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
santiph: Oh, you're right. I did assume the original spy has been set through Jasmine's spy. Now I see your point.
Regarding the library... I like the idea! Let's talk more about it on IRC or hangouts

assets/app/features/upgrade/controllers/upgrade-landing.controller.js
santiph: `upgrade-status.factory.js` looks like a good place to store this

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
skazi0: And same answer. This would be refactoring of multiple tests. A separate PR would be needed for that.

santiph: Please, create a tech debt and add a link to be tracked properly.

skazi0: backlog item created

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
santiph: I just noticed the recommendation is to use Sinon for spying: https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#stubbing-and-spying

assets/app/data/crowbar/services/upgrade-status.factory.spec.js
```diff
@@ -131,6 +136,77 @@ describe('Upgrade Status Factory', function () {
...
+
+                    $rootScope.$digest();
+                });
+
+                it('should call completed callback', function () {
+                    expect(mockedCompletedCallback).toHaveBeenCalled();
+                });
+
+                it('should not call running callback', function () {
+                    expect(mockedRunningCallback).not.toHaveBeenCalled();
```
santiph, line 168: Is it possible to assert the number of times both callbacks have been called?

santiph: Same for assertion on line 192

santiph:
Unit tests for `upgrade-upgrade-administration-server.controller.js` needs to be updated

assets/app/data/crowbar/services/upgrade-status.factory.spec.js
```diff
@@ -131,6 +136,77 @@ describe('Upgrade Status Factory', function () {
...
+        describe('syncStatusFlags function', function () {
+            it('should be defined', function () {
+                should.exist(upgradeStatusFactory.syncStatusFlags);
+                expect(upgradeStatusFactory.syncStatusFlags).toEqual(jasmine.any(Function));
+            });
+
+            describe('when received status is completed', function () {
+                beforeEach(function () {
+                    bard.mockService(upgradeFactory, {
+                        getStatus: $q.when(completedUpgradeResponse)
```
santiph, line 151: I'm not able to comment on line 6, but after talking with @MaximilianMeister its value should be `database` and not `admin_upgrade` (given the current status of the steps object).
I'm not sure how this would affect the rest of the tests.

assets/app/data/crowbar/services/upgrade-status.factory.spec.js
```diff
@@ -131,6 +136,77 @@ describe('Upgrade Status Factory', function () {
...
+
+                    $rootScope.$digest();
+                });
+
+                it('should call completed callback', function () {
+                    expect(mockedCompletedCallback).toHaveBeenCalledTimes(1);
+                });
+
+                it('should not call running callback', function () {
+                    expect(mockedRunningCallback).not.toHaveBeenCalledTimes(1);
```
santiph: If `onRunning` is executed none, 2 or more times, this assertion will be correct. Please fix it

assets/app/data/crowbar/services/upgrade-status.factory.spec.js
```diff
@@ -131,6 +136,77 @@ describe('Upgrade Status Factory', function () {
...
+
+                    $rootScope.$digest();
+                });
+
+                it('should not call completed callback', function () {
+                    expect(mockedCompletedCallback).not.toHaveBeenCalled();
+                });
+
+                it('should call running callback', function () {
+                    expect(mockedRunningCallback).toHaveBeenCalled();
```
santiph: `.toHaveBeenCalledTimes(1)`

assets/app/features/upgrade/controllers/upgrade-landing.controller.spec.js
```diff
@@ -3,6 +3,135 @@
   crowbarFactory $q $rootScope module $state */
 describe('Upgrade Landing Controller', function() {
     var controller,
+        completedUpgradeResponseData = {
+            current_step: 'upgrade_prepare',
```
MaximilianMeister: current_step should be `admin_backup` here, as `upgrade_prepare` is already passed

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.spec.js
```diff
@@ -175,22 +29,10 @@ describe('Upgrade Flow - Upgrade Administration Server Controller', function ()
...
-
-        it('should set completed to true if upgrade is completed', function () {
-            activeStatusResponse.data = completedUpgradeResponseData;
-            controller = $controller('UpgradeUpgradeAdministrationServerController');
-            $rootScope.$digest();
-            assert.isTrue(controller.adminUpgrade.completed);
-            assert.isFalse(controller.adminUpgrade.running);
+        it('should call syncStatusFlags() to update the state', function () {
+            expect(upgradeStatusFactory.syncStatusFlags).toHaveBeenCalledWith(
+                'admin_upgrade', controller.adminUpgrade, jasmine.any(Function), jasmine.any(Function)
```
santiph: Can you assert the functions that are being sent as parameters here?
`controller.waitForUpgradeToEnd` and `upgradeStepsFactory.setCurrentStepCompleted`? (I think you'd need to mock the service with a method called `setCurrentStepCompleted` and inject it to the controller on line 17 to be reused here)

santiph (changes requested):
A few more changes requested

assets/app/data/crowbar/services/upgrade-status.factory.spec.js
skazi0: Fixed. This should not affect tests.

assets/app/data/crowbar/services/upgrade-status.factory.spec.js
skazi0: Yeah, I changed wrong test.

assets/app/data/crowbar/services/upgrade-status.factory.spec.js
skazi0: ditto

assets/app/features/upgrade/controllers/upgrade-upgrade-administration-server.controller.spec.js
skazi0: `waitForUpgradeToEnd` is not used outside of the controller. I don't like exposing things just for tests.
The second part is done.

santiph (approved):

PR #97, opened 2016-11-30
upgrade: Restore last step on refresh or reopen

skazi0:
When browser window is closed and reopened or refreshed during upgrade flow,
the user should be presented with the most appropriate view of the flow.
The UI will show the page respective for `current_step` as returned by the
status API and/or state stored in browser.

Same mechanism is used to prevent user from jumping to any page in the flow
by manually entering the page URL into the address bar.

Note: this PR is just implementation of the general mechanism. Proper handling of page refresh needs to be implemented for each page (except Admin Server Upgrade which was implemented before).
Refresh handling is done in following PRs: https://github.com/crowbar/crowbar-ui/pull/95, ...

### DIFF
--- a/assets/app/crowbar-app.module.js
+++ b/assets/app/crowbar-app.module.js
@@ -21,6 +21,7 @@
             'crowbarApp.upgrade'
         ])
         .config(configuration)
+        .run(run)
         // lodash support
         .constant('_', _);
 
@@ -35,5 +36,28 @@
             // For any unmatched url, redirect to /upgrade/prepare
             $urlRouterProvider.otherwise('/upgrade/landing');
         }
+    }
+
+    run.$inject = ['$rootScope', '$state', 'upgradeFactory', 'upgradeStepsFactory'];
+
+    function run($rootScope, $state, upgradeFactory, upgradeStepsFactory) {
+        var cleanup = $rootScope.$on('$stateChangeStart',
+                function(event, toState/*, toParams, fromState, fromParams*/) {
+                    upgradeFactory.getStatus()
+                        .then(
+                            function (response) {
+                                var expectedState = upgradeStepsFactory.lastStateForRestore(response.data);
+
+                                if (toState.name !== expectedState) {
+                                    event.preventDefault();
+                                    $state.go(expectedState)
+                                        .then(function () { upgradeStepsFactory.refeshStepsList(); });
+                                }
+                            },
+                            function (/*errorResponse*/) {
+                            }
+                        );
+                });
+        $rootScope.$on('$destroy', cleanup);
     }
 })(_);
--- a/assets/app/crowbar-app.module.js
+++ b/assets/app/crowbar-app.module.js
@@ -21,7 +21,6 @@
             'crowbarApp.upgrade'
         ])
         .config(configuration)
-        .run(run)
         // lodash support
         .constant('_', _);
 
@@ -36,28 +35,5 @@
             // For any unmatched url, redirect to /upgrade/prepare
             $urlRouterProvider.otherwise('/upgrade/landing');
         }
-    }
-
-    run.$inject = ['$rootScope', '$state', 'upgradeFactory', 'upgradeStepsFactory'];
-
-    function run($rootScope, $state, upgradeFactory, upgradeStepsFactory) {
-        var cleanup = $rootScope.$on('$stateChangeStart',
-                function(event, toState/*, toParams, fromState, fromParams*/) {
-                    upgradeFactory.getStatus()
-                        .then(
-                            function (response) {
-                                var expectedState = upgradeStepsFactory.lastStateForRestore(response.data);
-
-                                if (toState.name !== expectedState) {
-                                    event.preventDefault();
-                                    $state.go(expectedState)
-                                        .then(function () { upgradeStepsFactory.refeshStepsList(); });
-                                }
-                            },
-                            function (/*errorResponse*/) {
-                            }
-                        );
-                });
-        $rootScope.$on('$destroy', cleanup);
     }
 })(_);
--- a/assets/app/features/upgrade/steps.factory.js
+++ b/assets/app/features/upgrade/steps.factory.js
@@ -145,7 +145,6 @@
         /**
          * Handler for ui-router state change events to ensure requested page matches current backend status
          * For parameter descriptions see: https://github.com/angular-ui/ui-router/wiki#state-change-events
-         * @return {string}
          */
         function validateRequestedState(event, toState/*, toParams, fromState, fromParams*/) {
             upgradeFactory.getStatus()
--- a/assets/app/features/upgrade/steps.factory.js
+++ b/assets/app/features/upgrade/steps.factory.js
@@ -4,14 +4,17 @@
         .module('crowbarApp.upgrade')
         .factory('upgradeStepsFactory', upgradeStepsFactory);
 
-    upgradeStepsFactory.$inject = ['$state', 'STEP_STATES', 'UPGRADE_LAST_STATE_KEY'];
+    upgradeStepsFactory.$inject = ['$state', 'upgradeFactory', 'STEP_STATES', 'UPGRADE_LAST_STATE_KEY'];
     /* @ngInject */
-    function upgradeStepsFactory($state, STEP_STATES, UPGRADE_LAST_STATE_KEY) {
+    function upgradeStepsFactory($state, upgradeFactory, STEP_STATES, UPGRADE_LAST_STATE_KEY) {
         var factory = {
             steps: initialSteps(),
             stepByState: stepByState,
             stepByID: stepByID,
+            validateRequestedState: validateRequestedState,
             lastStateForRestore: lastStateForRestore,
+            showNextStep: showNextStep,
+            isLastStep: isLastStep,
             activeStep: {},
             refeshStepsList: refeshStepsList,
             setCurrentStepCompleted: setCurrentStepCompleted,
@@ -99,34 +102,99 @@
             ];
         }
 
+        /**
+         * Validate if the active step is the last avilable step
+         * @return boolean
+         */
+        function isLastStep() {
+            return factory.steps[factory.steps.length - 1] === factory.activeStep;
+        }
+
+        /**
+         * Move to the next available step
+         */
+        function showNextStep() {
+            // Only move forward if active step isn't last step available
+            if (factory.isLastStep()) {
+                return;
+            }
+
+            var nextState = factory.steps[factory.activeStep.id + 1].state;
+            // save new state in local storage for proper "restore last step" handling
+            localStorage.setItem(UPGRADE_LAST_STATE_KEY, nextState);
+
+            $state.go(nextState);
+        }
+
+        /**
+         * Return step data for given state
+         */
         function stepByState(state) {
             return _.find(factory.steps, function (step) { return step.state === state; });
         }
 
+        /**
+         * Return step data for given id
+         */
         function stepByID(id) {
             return _.find(factory.steps, function (step) { return step.id === id; });
         }
 
+        /**
+         * Handler for ui-router state change events to ensure requested page matches current backend status
+         */
+        function validateRequestedState(event, toState/*, toParams, fromState, fromParams*/) {
+            upgradeFactory.getStatus()
+                .then(
+                    function (response) {
+                        var expectedState = factory.lastStateForRestore(response.data);
+
+                        if (toState.name !== expectedState) {
+                            event.preventDefault();
+                            $state.go(expectedState)
+                                .then(function () { factory.refeshStepsList(); });
+                        }
+                    },
+                    function (/*errorResponse*/) {
+                    }
+                );
+        }
+
+        /**
+         * For given `map`, return keys mapping to `wantedValue`
+         */
+        function keysForValue(map, wantedValue) {
+            var keys = [];
+            _.each(map, function (value, key) {
+                if (value === wantedValue) {
+                    keys.push(key);
+                }
+            });
+            return keys;
+        }
+
+        /**
+         * Return ui-router state appropriate for current backend status
+         */
         function lastStateForRestore(statusData) {
             var stepToStateMap = {
-                'upgrade_prechecks': 'upgrade-landing',
-                'upgrade_prepare': 'upgrade-landing',
-                'admin_backup': 'upgrade.backup',
-                'admin_repo_checks': 'upgrade.administration-repository-checks',
-                'admin_upgrade': 'upgrade.upgrade-administration-server',
-                'database': 'upgrade.database-configuration',
-                'nodes_repo_checks': 'upgrade.nodes-repositories-checks',
-                'nodes_services': 'upgrade.openstack-services',
-                'nodes_db_dump': 'upgrade.openstack-services',
-                'nodes_upgrade': 'upgrade.upgrade-nodes',
-                'finished': 'TODO(skazi): no UI for this step yet',
-            };
-
-            var currentStep = statusData.current_step,
+                    'upgrade_prechecks': 'upgrade-landing',
+                    'upgrade_prepare': 'upgrade-landing',
+                    'admin_backup': 'upgrade.backup',
+                    'admin_repo_checks': 'upgrade.administration-repository-checks',
+                    'admin_upgrade': 'upgrade.upgrade-administration-server',
+                    'database': 'upgrade.database-configuration',
+                    'nodes_repo_checks': 'upgrade.nodes-repositories-checks',
+                    'nodes_services': 'upgrade.openstack-services',
+                    'nodes_db_dump': 'upgrade.openstack-services',
+                    'nodes_upgrade': 'upgrade.upgrade-nodes',
+                    'finished': 'upgrade.upgrade-nodes',
+                },
+                currentStep = statusData.current_step,
                 steps = statusData.steps,
                 currentState = stepToStateMap[currentStep];
 
-            // select previous step if current is still pending
+            // select previous step if current is still pending and previous step has a different page
             if (steps[currentStep].status === STEP_STATES.pending) {
                 var storedLastState = localStorage.getItem(UPGRADE_LAST_STATE_KEY);
 
@@ -134,6 +202,13 @@
                 // show this instead of the previous one, otherwise apply the logic below to find
                 // proper page to show
                 if (storedLastState === currentState) {
+                    return currentState;
+                }
+
+                var backendStepsWithCurrentState = keysForValue(stepToStateMap, currentState);
+                // if there are more "backend" steps matching currentState, only move back if currentStep
+                // is the first one on the list
+                if (backendStepsWithCurrentState.length > 1 && backendStepsWithCurrentState[0] !== currentStep) {
                     return currentState;
                 }
 
--- a/assets/app/features/upgrade/steps.factory.js
+++ b/assets/app/features/upgrade/steps.factory.js
@@ -4,9 +4,9 @@
         .module('crowbarApp.upgrade')
         .factory('upgradeStepsFactory', upgradeStepsFactory);
 
-    upgradeStepsFactory.$inject = ['$state', 'upgradeFactory', 'STEP_STATES', 'UPGRADE_LAST_STATE_KEY'];
+    upgradeStepsFactory.$inject = ['$state', 'upgradeFactory', 'UPGRADE_STEP_STATES', 'UPGRADE_LAST_STATE_KEY'];
     /* @ngInject */
-    function upgradeStepsFactory($state, upgradeFactory, STEP_STATES, UPGRADE_LAST_STATE_KEY) {
+    function upgradeStepsFactory($state, upgradeFactory, UPGRADE_STEP_STATES, UPGRADE_LAST_STATE_KEY) {
         var factory = {
             steps: initialSteps(),
             stepByState: stepByState,
@@ -104,7 +104,7 @@
 
         /**
          * Validate if the active step is the last avilable step
-         * @return boolean
+         * @return {Boolean}
          */
         function isLastStep() {
             return factory.steps[factory.steps.length - 1] === factory.activeStep;
@@ -128,6 +128,7 @@
 
         /**
          * Return step data for given state
+         * @return {string}
          */
         function stepByState(state) {
             return _.find(factory.steps, function (step) { return step.state === state; });
@@ -135,6 +136,7 @@
 
         /**
          * Return step data for given id
+         * @return {string}
          */
         function stepByID(id) {
             return _.find(factory.steps, function (step) { return step.id === id; });
@@ -142,6 +144,8 @@
 
         /**
          * Handler for ui-router state change events to ensure requested page matches current backend status
+         * For parameter descriptions see: https://github.com/angular-ui/ui-router/wiki#state-change-events
+         * @return {string}
          */
         function validateRequestedState(event, toState/*, toParams, fromState, fromParams*/) {
             upgradeFactory.getStatus()
@@ -162,6 +166,7 @@
 
         /**
          * For given `map`, return keys mapping to `wantedValue`
+         * @return {Array} List of keys filtered by value
          */
         function keysForValue(map, wantedValue) {
             var keys = [];
@@ -175,6 +180,8 @@
 
         /**
          * Return ui-router state appropriate for current backend status
+         * @param {Object} Status data received from status API
+         * @return {string}
          */
         function lastStateForRestore(statusData) {
             var stepToStateMap = {
@@ -195,7 +202,7 @@
                 currentState = stepToStateMap[currentStep];
 
             // select previous step if current is still pending and previous step has a different page
-            if (steps[currentStep].status === STEP_STATES.pending) {
+            if (steps[currentStep].status === UPGRADE_STEP_STATES.pending) {
                 var storedLastState = localStorage.getItem(UPGRADE_LAST_STATE_KEY);
 
                 // if there is some stored state and it's pointing to page for current pending step
--- a/assets/app/features/upgrade/steps.factory.js
+++ b/assets/app/features/upgrade/steps.factory.js
@@ -4,11 +4,14 @@
         .module('crowbarApp.upgrade')
         .factory('upgradeStepsFactory', upgradeStepsFactory);
 
-    upgradeStepsFactory.$inject = ['$state'];
+    upgradeStepsFactory.$inject = ['$state', 'STEP_STATES', 'UPGRADE_LAST_STATE_KEY'];
     /* @ngInject */
-    function upgradeStepsFactory($state) {
+    function upgradeStepsFactory($state, STEP_STATES, UPGRADE_LAST_STATE_KEY) {
         var factory = {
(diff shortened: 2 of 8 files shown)


skazi0:

assets/app/crowbar-app.module.js
```diff
@@ -36,4 +37,27 @@
             $urlRouterProvider.otherwise('/upgrade/landing');
         }
     }
+
+    run.$inject = ['$rootScope', '$state', 'upgradeFactory', 'upgradeStepsFactory'];
+
+    function run($rootScope, $state, upgradeFactory, upgradeStepsFactory) {
```
santiph: Please, document what's the purpose of this run block.

assets/app/features/upgrade/steps.factory.js
```diff
@@ -95,5 +98,66 @@
                 }
             ];
         }
+
+        function stepByState(state) {
+            return _.find(factory.steps, function (o) { return o.state === state; });
```
santiph: What is `o`?

assets/app/features/upgrade/upgrade.constants.js
```diff
@@ -12,6 +12,7 @@
             'ha': ['clusters_healthy'],
             'ceph': ['ceph_healthy']
         })
+        .constant('UPGRADE_LAST_STATE_KEY', 'soc.upgrade.lastUIState')
```
santiph: What does this key means?

assets/app/features/upgrade/steps.factory.js
```diff
@@ -95,5 +98,66 @@
...
+                'admin_upgrade': 'upgrade.upgrade-administration-server',
+                'database': 'upgrade.database-configuration',
+                'nodes_repo_checks': 'upgrade.nodes-repositories-checks',
+                'nodes_services': 'upgrade.openstack-services',
+                'nodes_db_dump': 'upgrade.openstack-services',
+                'nodes_upgrade': 'upgrade.upgrade-nodes',
+                'finished': 'TODO(skazi): no UI for this step yet',
+            };
+
+            var currentStep = statusData.current_step,
```
santiph: `var` statement could be avoided

assets/app/features/upgrade/steps.factory.js
```diff
@@ -95,5 +98,66 @@
...
+        function stepByState(state) {
+            return _.find(factory.steps, function (o) { return o.state === state; });
+        }
+
+        function stepByID(id) {
+            return _.find(factory.steps, function (o) { return o.id === id; });
+        }
+
+        function lastStateForRestore(statusData) {
+            var mapping = {
```
santiph: Could this variable name be more specific?

assets/app/features/upgrade/steps.factory.js
```diff
@@ -95,5 +98,66 @@
...
+                var storedLastState = localStorage.getItem(UPGRADE_LAST_STATE_KEY);
+
+                // if there is some stored state and it's pointing to page for current pending step
+                // show this instead of the previous one, otherwise apply the logic below to find
+                // proper page to show
+                if (storedLastState === currentState) {
+                    return currentState;
+                }
+
+                var currentStepData = stepByState(currentState);
```
santiph, line 221: Please, declare all variables together and initialise them as needed

assets/app/crowbar-app.module.js
santiph: Also, how could this be tested?
According to John Papa, a factory could be used to solve the problem: https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#run-blocks

santiph (changes requested):
Partially reviewed, but looks like a good implementation overall. 
I'd consider https://www.npmjs.com/package/angular-local-storage that has an automatic fallback to cookies.

Itxaka:
Looks good for the moment, barring Santiago's comments.

assets/app/crowbar-app.module.js
skazi0: Implementing the main functionality in a service was exactly what I did here. I didn't want to imeplement `$on('$stateChangeStart')` handler in the factory though as this would require injecting many additional things there.
I thought I could make `lastStepForRestore()` function parameter-less but the status data is received in an async way so it would require some promise magic to make this work. This is why I did the `getStatus()` part on the app module side.
Note, that the main (and most complicated one) logic which requires testing is enclosed in `lastStepForRestore()` which is a pure function (except the localStorage part) and as such can be easily tested without any mocking etc.

skazi0:
As for using an angular wrapper around `localStorage`, I would prefer https://github.com/gsklee/ngStorage (check their arguments on not having cookie fallback).
However in this particular case I don't think it makes sense. The main goal of those wrappers is to provide angular bindings of models to `localStorage` items which is not needed in our scenarios. We need to have fine-grained control on when the values are saved and loaded to make everything work as expected.
Using a library in this case would add another dependency while not providing any benefit as we won't be using most of it's functionality anyway.

assets/app/features/upgrade/steps.factory.js
skazi0: If you mean using `statusData.current_step` directly then I don't agree. The "step" vs "state" naming is confusing enough for someone reading the code so I'd like to limit the complexity of statements below by clear and consistent naming of things.

assets/app/features/upgrade/steps.factory.js
skazi0: Why would I do that? It's not C. See: http://stackoverflow.com/questions/5053073/is-defining-every-variable-at-the-top-always-the-best-approach

assets/app/features/upgrade/upgrade.constants.js
skazi0: It's a key under which last upgrade state will be stored in localStorage. Those are scoped to the protocol+domain context so we need to make sure it doesn't cause conflicts with other parts of application.

assets/app/features/upgrade/steps.factory.js
skazi0: `o` stands for object but I changed it to `step`.

assets/app/features/upgrade/steps.factory.js
skazi0: renamed.

assets/app/crowbar-app.module.js
```diff
@@ -21,6 +21,7 @@
             'crowbarApp.upgrade'
         ])
         .config(configuration)
+        .run(run)
```
santiph: Please move this implementation to `crowbarApp.upgrade` module. `crowbarApp` is intended to be a wider module for other features than the upgrade process

Itxaka:
Looks good to me on a logical side. Technically, I dont think I cant comment as my js knowledge is not that good as to weight in.

assets/app/features/upgrade/steps.factory.js
```diff
@@ -95,5 +98,66 @@
                 }
             ];
         }
+
+        function stepByState(state) {
```
santiph, line 133: DocBlock please for `stepByState`, `stepByID` and `lastStateForRestore`

assets/app/features/upgrade/steps.factory.js
santiph: nono, I mean you could've used the same `var` from line 111

```
            var stepToStateMap = {
                 'upgrade_prechecks': 'upgrade-landing',
                 ...
                 'finished': 'TODO(skazi): no UI for this step yet',
             },
             currentStep = statusData.current_step,
```

assets/app/features/upgrade/upgrade.constants.js
santiph: `Suse Openstack Cloud`
In order to avoid abbreviations and long names, would you consider `crowbar.upgrade.lastUIState` as an option?

assets/app/features/upgrade/steps.factory.js
```diff
@@ -95,5 +98,66 @@
...
+                'upgrade_prepare': 'upgrade-landing',
+                'admin_backup': 'upgrade.backup',
+                'admin_repo_checks': 'upgrade.administration-repository-checks',
+                'admin_upgrade': 'upgrade.upgrade-administration-server',
+                'database': 'upgrade.database-configuration',
+                'nodes_repo_checks': 'upgrade.nodes-repositories-checks',
+                'nodes_services': 'upgrade.openstack-services',
+                'nodes_db_dump': 'upgrade.openstack-services',
+                'nodes_upgrade': 'upgrade.upgrade-nodes',
+                'finished': 'TODO(skazi): no UI for this step yet',
```
santiph: I think it'll be `nodes_upgrade`

assets/app/features/upgrade/steps.factory.js
```diff
@@ -95,5 +98,66 @@
...
+        function stepByState(state) {
+            return _.find(factory.steps, function (step) { return step.state === state; });
+        }
+
+        function stepByID(id) {
+            return _.find(factory.steps, function (step) { return step.id === id; });
+        }
+
+        function lastStateForRestore(statusData) {
+            var stepToStateMap = {
```
santiph, line 186: Shouldn't this map be stored as a constant? That way it'd be easier to reuse it in the unit test

assets/app/features/upgrade/upgrade.controller.js
```diff
@@ -40,7 +50,11 @@
                 return;
             }
 
-            $state.go(vm.steps.list[upgradeStepsFactory.activeStep.id + 1].state);
+            var nextState = vm.steps.list[upgradeStepsFactory.activeStep.id + 1].state;
+            // save new state in local storage for proper "restore last step" handling
+            localStorage.setItem(UPGRADE_LAST_STATE_KEY, nextState);
```
santiph: What do you think of using the steps.factory to handle the last 3 statements?

assets/app/features/upgrade/upgrade.controller.js
```diff
@@ -56,6 +70,8 @@
          */
         function cancelUpgrade() {
             upgradeFactory.cancelUpgrade().then(function(/* response */) {
+                // clear last page seen by the user
+                localStorage.removeItem(UPGRADE_LAST_STATE_KEY);
```
santiph, line 49: Same comment as line 55

assets/app/features/upgrade/upgrade.controller.js
santiph: Also, unit tests needs to be updated

assets/app/features/upgrade/steps.factory.js
santiph: In order to keep consistency with the rest of the application. Except for specific control structures, variables tend to be defined at the beginning of the functions.
But I just realised this is only used in the next condition only.

assets/app/features/upgrade/steps.factory.spec.js
```diff
@@ -50,12 +50,292 @@ describe('Stepes Factory', function () {
...
             finished: false
         }
-    ];
+        ],
+        initialStatusData = {
+            current_step: 'upgrade_prechecks',
+            substep: null,
+            current_node: null,
+            steps: {
+                upgrade_prechecks: {
```
santiph, line 59: This is not the first test I see with this set of configuration... Would you consider creating a helper for the unit tests to be reused across several ones?

santiph: Check https://github.com/johnpapa/generator-hottowel/blob/master/app/templates/src/client/test-helpers/mock-data.js
We should probably need to make a filter for `angularFeatures` gulp task defined in `gulp/angular.js` so we don't ship it to the client... but it could make test easier, shorter and more readable

assets/app/features/upgrade/steps.factory.spec.js
```diff
@@ -1,4 +1,4 @@
-/*global bard should assert expect upgradeStepsFactory module $state*/
+/*global bard should assert expect upgradeStepsFactory module $state UPGRADE_LAST_STATE_KEY */
 describe('Stepes Factory', function () {
     var mockedInitialSteps = [
```
santiph, line 3: This collection is duplicated from steps.factory. Can be reused?

assets/app/features/upgrade/steps.factory.spec.js
```diff
@@ -118,5 +406,87 @@ describe('Stepes Factory', function () {
...
+                expect(upgradeStepsFactory.stepByID(9999)).not.toBeDefined();
+            });
+        });
+
+        describe('lastStateForRestore function', function () {
+            describe('when there is no stored last state', function () {
+                beforeEach(function () {
+                    localStorage.removeItem(UPGRADE_LAST_STATE_KEY);
+                });
+                it('should return "upgrade-landing" before any step', function() {
```
santiph, line 734: It would be interesting to have this set of validations be done by iterating through `stepsToStateMap` from the steps.factory and cover all cases, if such map is stored as a public property of the factory or a module's constant... What do you think?

assets/app/features/upgrade/steps.factory.spec.js
```diff
@@ -118,5 +406,87 @@ describe('Stepes Factory', function () {
...
+                it('should return "upgrade.backup" after backup', function() {
+                    expect(upgradeStepsFactory.lastStateForRestore(statusDataAfterBackup))
+                        .toEqual('upgrade.backup');
+                });
+                it('should return "upgrade.administration-repository-checks" after admin repo checks', function() {
+                    expect(upgradeStepsFactory.lastStateForRestore(statusDataAfterAdminRepoChecks))
+                        .toEqual('upgrade.administration-repository-checks');
+                });
+            });
+            describe('when stored state is invalid in given case', function () {
```
santiph: step vs state.
`upgrade-landing` is a valid state, but not a valid step in the upgrade wizard.

assets/app/features/upgrade/steps.factory.spec.js
```diff
@@ -118,5 +406,87 @@ describe('Stepes Factory', function () {
...
+                    localStorage.setItem(UPGRADE_LAST_STATE_KEY, 'upgrade-landing');
+                });
+                it('should return "upgrade.backup" after backup', function() {
+                    expect(upgradeStepsFactory.lastStateForRestore(statusDataAfterBackup))
+                        .toEqual('upgrade.backup');
+                });
+            });
+            describe('when stored state points to current, pending step', function () {
+                beforeEach(function () {
+                    localStorage.setItem(UPGRADE_LAST_STATE_KEY, 'upgrade.administration-repository-checks');
```
santiph, line 808: I'd like to see this validation for the other cases

santiph (changes requested):
Woahh! Very interesting PR. 
I'd like to get the card updated before moving forward so we're sure to implement the right requirements.

It's looking very good overall. Please check the comments.

Regarding the cookies fallback, I'd try to consider how many customers would use browsers like IE10* that doesn't support it and how the application will behave in those cases. If it'll throw an error and stop everything, then we need to implement some sort of "Unsupported browser, please upgrade" warning message (a new card needs to be created for that).

*IE9 or IE10 is shipped by default with Windows 7. Is there any still using something like that to use crowbar? Will they have to change their browser/OS just to have compatibility with localStorage and be able to run the upgrade?

assets/app/crowbar-app.module.js
santiph: Untested Cyclomatic complexity, if I counted right, is 5 for this function. Too high for something that will affect the entire upgrade flow.
Let's discuss other alternatives on monday

assets/app/features/upgrade/steps.factory.js
skazi0: True. I left it like this as I was considering moving the mapping somewhere else. I will remove the extra `var`.

assets/app/features/upgrade/steps.factory.js
skazi0: Yes, this is exactly why it was declared here. It is not used above and if we leave it uninitialized but declared above it would make it more confusing if someone needs to change the code later.

assets/app/features/upgrade/upgrade.constants.js
skazi0: OK.

skazi0:
Regarding `localStorage`, check this: https://developer.mozilla.org/en/docs/Web/API/Window/localStorage#Browser_compatibility. TLDR: it's supported since IE8.

assets/app/features/upgrade/steps.factory.js
skazi0: I don't think using this mapping in the unit tests is a good idea as this will hide errors in the mapping itself. The unit tests should explicitly contain input and output values for such cases.

assets/app/features/upgrade/steps.factory.js
skazi0: You mean that the page for `finished` will be the same as for `nodes_upgrade` (i.e. `upgrade.upgrade-nodes`)?

assets/app/features/upgrade/upgrade.controller.js
skazi0: Good idea.

assets/app/features/upgrade/steps.factory.js
santiph: Once the nodes upgrade is completed, the user will see a "Finish" button in the navbar (instead of "Next"). On click, will be redirected (still under discussion) to crowbar dashboard or barclamps page if couldn't be reapplied automatically.

So I guess the state is `upgrade.upgrade-nodes` for this case.

santiph:
My bad. I used http://caniuse.com/#search=localstorage to check about localstorage and I have misinterpreted the information there.

assets/app/features/upgrade/upgrade.controller.js
skazi0: I'm not sure about this one...

assets/app/features/upgrade/steps.factory.spec.js
skazi0: I don't like creating dependencies between unit tests. We have big test inputs here and we have to live with this :(

assets/app/features/upgrade/steps.factory.spec.js
skazi0: This doesn't fit into the scope of this PR.

assets/app/features/upgrade/steps.factory.spec.js
skazi0: As mentioned above, I believe that unit tests should not rely on the mappings from the tested module. They should not test if "the module does what the module does" but rather if "the module does as explicitly specified".

assets/app/features/upgrade/steps.factory.spec.js
skazi0: There's nothing about step here. `upgrade-landing` (stored in `localStorage`) is invalid in this test as it doesn't match the status data passed to `lastStateForRestore()`. In such cases it will be ignored and data from function argument will be used.

assets/app/features/upgrade/steps.factory.spec.js
skazi0: I added more validations to the section without localStorage part. These are supposed to ensure the mapping is correct.

assets/app/features/upgrade/upgrade.module.js
```diff
@@ -2,5 +2,14 @@
     'use strict';
 
     angular
-        .module('crowbarApp.upgrade', ['ui.router', 'ngFileSaver', 'pascalprecht.translate', 'suseData']);
+        .module('crowbarApp.upgrade', ['ui.router', 'ngFileSaver', 'pascalprecht.translate', 'suseData'])
+        .run(run);
+
+    run.$inject = ['$rootScope', 'upgradeStepsFactory'];
+
+    function run($rootScope, upgradeStepsFactory) {
```
santiph, line 14: Missing docblock

assets/app/features/upgrade/steps.factory.js
```diff
@@ -95,5 +101,138 @@
...
+            var nextState = factory.steps[factory.activeStep.id + 1].state;
+            // save new state in local storage for proper "restore last step" handling
+            localStorage.setItem(UPGRADE_LAST_STATE_KEY, nextState);
+
+            $state.go(nextState);
+        }
+
+        /**
+         * Return step data for given state
+         */
```
santiph, line 132: missing `@return`

assets/app/features/upgrade/steps.factory.js
```diff
@@ -95,5 +101,138 @@
...
+        /**
+         * Return step data for given id
+         */
+        function stepByID(id) {
+            return _.find(factory.steps, function (step) { return step.id === id; });
+        }
+
+        /**
+         * Handler for ui-router state change events to ensure requested page matches current backend status
+         */
```
santiph, line 148: Please, document parameters

assets/app/features/upgrade/steps.factory.js
```diff
@@ -95,5 +101,138 @@
...
+                }
+
+                var currentStepData = stepByState(currentState);
+
+                // getting `undefined` from `stepByState()` should happen only
+                // for landing page which is not included in the wizard UI.
+                if (angular.isDefined(currentStepData)) {
+                    if (currentStepData.id > 0) {
+                        return stepByID(currentStepData.id - 1).state;
+                    } else {
```
santiph: please remove the `else` section if not needed

assets/app/features/upgrade/steps.factory.js
```diff
@@ -95,5 +101,138 @@
...
+                        // don't move user back from backup page to landing page
+                        // this would require another "prepare" call to move forward
+                    }
+                } else if (currentState === 'upgrade-landing'){
+                    // this is the only valid case when expectedStep can be undefined
+                    // expectedState is already on landing page so nothing needs to be done
+                    // keeping this empty block here to make logic clear
```
santiph, line 235: Please avoid empty blocks. Use a comment to clarify conditions instead

assets/app/features/upgrade/steps.factory.js
```diff
@@ -95,5 +101,138 @@
...
+                if (value === wantedValue) {
+                    keys.push(key);
+                }
+            });
+            return keys;
+        }
+
+        /**
+         * Return ui-router state appropriate for current backend status
+         */
```
santiph, line 184: missing parameters and return sections

assets/app/features/upgrade/steps.factory.spec.js
santiph: If a different state has been stored than the one informed by the service, what should `lastStateForRestore` return?

assets/app/features/upgrade/steps.factory.spec.js
```diff
@@ -50,13 +50,541 @@ describe('Stepes Factory', function () {
             active: false,
             finished: false
         }
-    ];
+        ],
+        initialStatusData = {
```
santiph, line 54: There are 13 exact same structures hardcoded here with just different values occupying 500+ lines of code. There must be a more efficient and maintainable way to generate this.

santiph (changes requested):
Left some more comments.
Please, double check the functions created and complete their docblocks

assets/app/features/upgrade/steps.factory.js
skazi0: Where would I put the comment if there's no block which this comment applies to? I could invert the logic and avoid empty blocks but it would make things very hard to understand for someone reading the code later.

assets/app/features/upgrade/steps.factory.js
skazi0: Same as below regarding comments in empty blocks.

assets/app/features/upgrade/steps.factory.spec.js
skazi0: Do you want to write code to generate test fixtures? How are you going to test this code?

assets/app/features/upgrade/steps.factory.spec.js
skazi0: This is covered exactly by "when stored state is invalid in given case" test.

assets/app/features/upgrade/upgrade.controller.js
santiph: Yes? No? Why?

assets/app/features/upgrade/steps.factory.js
santiph: A comment at the beginning of the conditional section, or even at the end explaining the remaining cases should be enough and will also help to reduce the complexity of this method

assets/app/features/upgrade/steps.factory.spec.js
santiph: Follow up card: https://trello.com/c/ZOJZdSuH/2120-research-options-for-efficient-unit-testing-of-responses-from-status-api

assets/app/features/upgrade/steps.factory.spec.js
santiph: Follow up card: https://trello.com/c/ZOJZdSuH/2120-research-options-for-efficient-unit-testing-of-responses-from-status-api

assets/app/features/upgrade/steps.factory.spec.js
santiph: `upgrade-landing` is a valid state, not step (which is confusing enough, but still valid for that case), but it doesn't validate `lastStateForRestore` method returns when a different (yet valid) state is stored in localstorage.
A case for this could be if the user decides to open a different browser to continue the upgrade flow some steps and then refresh the original one, or any error that could've been introduced in `showNextStep` or `cancelUpgrade` methods in the factory and upgrade.controller

skazi0: This is exactly as I said. "invalid in given case" means "not in sync with backend".

skazi0: Feel free to propose different wording if you think this one is not clear.

assets/app/features/upgrade/upgrade.controller.js
skazi0: The other case already used internals of `upgradeStepsFatory` which seemed out of place. Here we have something which is very specific to cancel feature and will likely change in the future. I would leave it like this for now.

assets/app/features/upgrade/steps.factory.js
skazi0: If you're so worried about the computer wasting time running those empty blocks I can enclose them in comments together with the explanation. Plain descriptions are not enough IMO as they will not have anything to refer to.

assets/app/features/upgrade/steps.factory.js
```diff
@@ -95,5 +101,145 @@
...
+         * @return {string}
+         */
+        function stepByID(id) {
+            return _.find(factory.steps, function (step) { return step.id === id; });
+        }
+
+        /**
+         * Handler for ui-router state change events to ensure requested page matches current backend status
+         * For parameter descriptions see: https://github.com/angular-ui/ui-router/wiki#state-change-events
+         * @return {string}
```
santiph: This method is not returning a string

assets/app/features/upgrade/upgrade.controller.js
santiph: Good point

santiph (approved):

PR #98, opened 2016-12-06
upgrade: Admin backup page refresh handling

skazi0:
When admin server backup was created before page was (re)opened,
Next button should be enabled to avoid need to create another
backup in order to continue with upgrade


santiph:
Please, double check with @kwwii.
I slightly remember the requirement is different and the user needs to create a new backup before continue, even when the action has been completed previously.

skazi0:

skazi0:
It was decided, that we don't want this functionality on the backup page. User can easily create another backup before going further.

PR #100, opened 2016-12-07
upgrade: more coverage for upgrade-status.factory.js

Itxaka:
Tests the actual calls to the functions instead of testing for the timeout to have been called. Uses $timeout.flush() so indirectly is testing that the timeout works.

### DIFF
--- a/assets/app/data/crowbar/services/upgrade-status.factory.spec.js
+++ b/assets/app/data/crowbar/services/upgrade-status.factory.spec.js
@@ -1,4 +1,4 @@
-/*global bard assert $q $rootScope should expect module upgradeFactory upgradeStatusFactory */
+/*global bard assert $q $rootScope should expect $timeout upgradeFactory upgradeStatusFactory */
 describe('Upgrade Status Factory', function () {
     var pollingInterval = 1234,
         testedStep = 'admin_upgrade',
@@ -107,7 +107,6 @@ describe('Upgrade Status Factory', function () {
         mockedErrorCallback,
         mockedRunningCallback,
         mockedCompletedCallback,
-        mockedTimeout,
         errorResponse = {
             error: 'some error'
         };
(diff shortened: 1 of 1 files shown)


skazi0 (changes requested):
I don't like this change at all... testing for `$timeout` call should be enough assuming that we can rely on external modules being tested.

The coverage increase is marginal and you loose important piece of test here. You only see that `waitForStepToEnd()` was called but you don't see if it was called via `$timeout`. Try removing the `$timeout` from the service code, call `waitForStepToEnd()` directly and the test will still pass. Current version will not behave too good either (infinite loop probably) but it will at least not hide the problem.

Itxaka:
@skazi0 maybe I can put a spy on $timeot and check if timeout was called as well?

skazi0:
`$timeout` is a tricky beast. Doesn't like being spied on :)

santiph:
As `$timeout` is an angular service and we can assume it works as expected.
We should test `$timeout` has been called as expected, but not its inner functionality nor assert the method is actually being called after the delay.

Itxaka:
> I don't like this change at all... testing for $timeout call should be enough assuming that we can rely on external modules being tested.

> We should test $timeout has been called as expected, but not its inner functionality nor assert the method is actually being called after the delay.

Not agreed.

Changing this line https://github.com/crowbar/crowbar-ui/blob/master/assets/app/data/crowbar/services/upgrade-status.factory.js#L64 to nonsense will make the tests pass with flying colours. The actual tests are also hiding part of the code working properly which could lead to issues in the future.


But I dont want to mock and reproduce the whole $timeout function to be able to spy on it, so I will close it. we will still have code in there *not* tested, but seems that we can live with that.¯\\\_(ツ)\_\/¯

skazi0:
I found a better way to solve this problem by changing the service itself. Turns out `$timeout` has an optional last parameter which can be used to pass arguments to the callback function. With this in place, testing becomes way easier.

New PR: https://github.com/crowbar/crowbar-ui/pull/103

PR #101, opened 2016-12-07
upgrade: increase coverage for upgrade.factory.js

Itxaka:


assets/app/data/crowbar/services/upgrade.factory.spec.js
```diff
@@ -9,13 +9,43 @@ describe('Upgrade Factory', function () {
...
+                'success': true
+            },
+            'schema_migration': {
+                'success': true
+            },
+            'crowbar_init': {
+                'success': true
+            }
+        },
+        mockedConnectDatabaseServerResponse = {
```
santiph: How is this different to `mockedNewDatabaseServerResponse` from line 12?

Itxaka: Its not, but for the cost of one var we can visually separate them in the tests for easier comprehension of what it refers to.

skazi0: I would change it to `mockedConnectDatabaseServerResponse = mockedNewDatabaseServerResponse`. This gives us the effect of separation as you mentioned but also shows that at the moment the responses are identical.

skazi0 (changes requested):
small nit as commented

assets/app/data/crowbar/services/upgrade.factory.spec.js
Itxaka: done!

Itxaka:
Need rebase after https://github.com/crowbar/crowbar-ui/pull/104 is in

skazi0 (approved):

santiph:
No issues found, approved

PR #105, opened 2016-12-14
upgrade: Removed unused dependency

skazi0:


Itxaka (approved):
Do we need @rsalevsky to rebuild the package so this gets dropped?

skazi0:
No, this was just missed injection. It doesn't change real package-level deps.

PR #107, opened 2016-12-15
upgrade: generic error handling

Itxaka:
Missing tests


Itxaka:
[(3) UI: Generic error handling (landing page)](https://trello.com/c/oo5N6XNL/204-3-ui-generic-error-handling-landing-page)

assets/app/widgets/suse-modal/suse-modal.directive.js
```diff
@@ -0,0 +1,27 @@
...
+            templateUrl: 'app/widgets/suse-modal/suse-modal.directive.html',
+            scope: {
+                error: '='
+            },
+            link: function (scope) {
+                scope.$watch('error', function(value) {
+                    if (angular.isDefined(value)) {
+                        scope.title = value.title;
+                        scope.body = value.body;
+                        $('#suseModal').modal('show');
```
skazi0: Why not work on the `elem` object passed to the link function here?

skazi0 (changes requested):

assets/app/widgets/suse-modal/suse-modal.directive.js
```diff
@@ -0,0 +1,32 @@
+(function() {
+    'use strict';
+
+    angular
+        .module('crowbarWidgets')
+        .directive('suseModal', suseModal);
+
+    function suseModal($uibModal) {
```
santiph, line 18: `suseModal.$inject = ['$uibModal'];`

assets/app/widgets/suse-modal/suse-modal.directive.js
```diff
@@ -0,0 +1,32 @@
...
+            restrict: 'E',
+            scope: {
+                error: '='
+            },
+            link: function (scope) {
+                scope.$watch('error', function(value) {
+                    if (angular.isDefined(value)) {
+                        $uibModal.open({
+                            templateUrl: 'app/widgets/suse-modal/suse-modal.directive.html',
+                            controller: function($uibModalInstance, $scope){
```
santiph: I think `$inject` will also be needed for this controller when minified

assets/app/widgets/suse-modal/suse-modal.directive.jade
```diff
@@ -0,0 +1,9 @@
+.modal-content
+    .modal-header
+        button.close(type='button', ng-click='dismiss()', aria-label='Close')
+            span(aria-hidden='true') ×
+        h4#suseModalLabel.modal-title {{ error.title }}
+    .modal-body
+        p {{ error.body }}
+    .modal-footer
+        button.btn.btn-default(type='button', ng-click='dismiss()') Close
```
santiph: missing new line at the end of the file

assets/app/widgets/suse-modal/suse-modal.directive.jade
```diff
@@ -0,0 +1,9 @@
+.modal-content
+    .modal-header
+        button.close(type='button', ng-click='dismiss()', aria-label='Close')
+            span(aria-hidden='true') ×
+        h4#suseModalLabel.modal-title {{ error.title }}
```
santiph: Please keep in mind that more than one suse-modal could be invoked in the same page. Is an Id going to be enough?
Also, one way data binding?

assets/app/features/upgrade/templates/landing-page.jade
```diff
@@ -11,3 +11,4 @@
     .row.hidden-md.hidden-lg.step-3(ng-class="{'enabled': upgradeLandingVm.mode.valid}")
         .col-md-6
             include landing-page_step3.partial.jade
+suse-modal(error="upgradeLandingVm.error")
```
santiph, line 14: I think the controller should be the one creating/deleting modals as error appears.

santiph (changes requested):

assets/app/features/upgrade/templates/landing-page.jade
skazi0: The idea is that the directive will encapsulate the modal. The controller will be just used to bind the directive with the error source. Controllers should not create/delete view elements.

assets/app/widgets/suse-modal/suse-modal.directive.jade
Itxaka: How is that gonna work? You can only show one modal per page (unless you want to show several at the same time, which is probably terrible UI/UX) and in every page that we have, we call something then continue or fail. Is in that fail when we show the modal with the appropiate error/errors aggregated. 


So basically check -> failure -> modal -> recheck and start again. 

I dont see where we could have more than one modal invoked. This way we reuse the same modal to push and show the error that we want to show.

assets/app/widgets/suse-modal/suse-modal.directive.js
Itxaka: Good point!

assets/app/widgets/suse-modal/suse-modal.directive.jade
santiph: `suse-modal` sounds generic enough to be also used by a login form, not just errors.
As a suse directive, it should be generic enough to be reused in other pages/projects/contexts, and there is my concern

assets/app/widgets/suse-modal/suse-modal.directive.js
Itxaka: also done!

assets/app/widgets/suse-modal/suse-modal.directive.jade
Itxaka: done

assets/app/widgets/suse-modal/suse-modal.directive.jade
Itxaka: Is that a real use case for this? Im pretty sure a login form will require much more code than what this is now. 

Lets think of the now instead of creating stuff to cover 99% of the use cases, because in the end those rarely happen. I can change the name to be upgradeModal if you want.

skazi0: I would rather call this something like `suse-error-modal` as this is the current use case.

skazi0 (approved):

assets/app/widgets/suse-modal/suse-modal.directive.jade
santiph: Missing translation

assets/app/widgets/suse-modal/suse-modal.directive.jade
Itxaka: just removed it, it was not used for anything anyway

santiph (changes requested):

assets/app/widgets/suse-modal/suse-modal.directive.jade
```diff
@@ -0,0 +1,9 @@
+.modal-content
+    .modal-header
+        button.close(type='button', ng-click='dismiss()', aria-label='Close')
+            span(aria-hidden='true') ×
+        h4.modal-title {{ error.title }}
+    .modal-body
+        p {{ error.body }}
+    .modal-footer
+        button.btn.btn-default(type='button', ng-click='dismiss()') Close
```
santiph: Missing translation

santiph (changes requested):

santiph (approved):

PR #113, opened 2017-01-03
upgrade: Stop waiting on failed step

skazi0:
Status API returns success HTTP code when current step has failed.
The actual status inside response needs to be checked to handle
such cases properly.
If that happens `onError` callback is triggered with the whole
response received from the API. Status code and response content
can be used to distinguish between different error cases.

BONUS: the tests were cleaned up a bit

### DIFF
--- a/assets/app/data/crowbar/services/upgrade-status.factory.js
+++ b/assets/app/data/crowbar/services/upgrade-status.factory.js
@@ -65,9 +65,13 @@
             upgradeFactory.getStatus()
                 .then(
                     function (response) {
+                        var stepStatus = response.data.steps[step].status;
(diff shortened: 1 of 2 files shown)


Itxaka (approved):
Looks good!

skazi0:
[(3) UI: Generic error handling (admin backup)](https://trello.com/c/KwAOzuC2/205-3-ui-generic-error-handling-admin-backup)

matelakat (approved):
Looks good to me. I think I would have preferred the incomplete -> partial refactor step to be in a different commit.

skazi0:
@matelakat good point, I split the changes into two commits.

PR #117, opened 2017-01-05
upgrade: Error modal common format

skazi0:
The error modal was modified to accept errors in common format returned from the API.
Landing page was updated to match this.

Errors passed to modal should be defined in following format:
```
errors = {
  title: "optional_title_code",
  errors: {
    error_code1: { data: [ 1, 2, 3 ], help: "some help text" },
    error_code2: { data: "cmd output text", help: "fix the problem" }
  }
}
```
All error codes and title are translated and scoped using `translation-prefix` directive attribute (default: `modal`).
All upgrade related error codes and messages should be placed in `upgrade.errors` sections of the language file.
If title is omitted, `default-title` code is used (with scoping prefix added).

For simple errors returned from the backend, this leads to basic implementation:
```
errors = response.data;
```

### DIFF
--- a/assets/app/features/upgrade/controllers/upgrade-landing.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-landing.controller.js
@@ -20,7 +20,8 @@
         'ADDONS_PRECHECK_MAP',
         'PREPARE_TIMEOUT_INTERVAL',
         'UPGRADE_STEPS',
-        'UPGRADE_MODES'
+        'UPGRADE_MODES',
+        'UNEXPECTED_ERROR_DATA',
     ];
     // @ngInject
     function UpgradeLandingController(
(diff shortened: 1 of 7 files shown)


skazi0:
zrzut ekranu z 2017-01-05 15-42-39
zrzut ekranu z 2017-01-09 10-24-12

Example precheck error on the landing page.

NOTES:
**1. overall styling of the error list needs to be defined/implemented by the UI team (scrollbar might be needed if there are many errors).**
2. red title is translated from the error code (key). other errors need to be added to the translations file.
3. the `data` field (first line below red title) is an array which is not properly formatted yet in the modal. backend can return different formats here (string, array, object) and proper formatting needs to be defined for these by the UI team.

skazi0:
[(3) UI: Generic error handling (admin backup)](https://trello.com/c/KwAOzuC2/205-3-ui-generic-error-handling-admin-backup)

Itxaka (approved):

skazi0:
@kwwii please take a look at the screenshots and think about some proper design for the "generic" display of errors.

PR #122, opened 2017-01-13
upgrade: Fix style of disabled tab

skazi0:
After database configuration is done, tabs in the UI are disabled
but out styles override default "disabled" look. This change adds
custom style for disabled tabs to restore the default look.


jsuchome (approved):
Tested with 

``
.upgrade-flow .database-configuration-page .nav-tabs > li.disabled > a {
  color: #777;
}
`` in app.css, looks good

Itxaka (approved):

PR #128, opened 2017-01-24
upgrade: Minor landing page improvements

skazi0:
Some missing error code descriptions were added. Display of errors
for non-required checks was fixed. Wording was improved in some of
the messages.

### DIFF
--- a/assets/app/features/upgrade/controllers/upgrade-landing.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-landing.controller.js
@@ -79,7 +79,7 @@
 
         vm.mode = {
             active: false,
-            type: false,
+            type: UPGRADE_MODES.none,
             valid: false
         };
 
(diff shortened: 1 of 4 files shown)


jsuchome:
I cannot comment on JS code but I'm glad to see those new error keys included. Thanks!

Itxaka:
Testing this locally, during the first click, some checks will fail

```
Network Health
Cloud Health
Storage Health
High Availability Health
```

And the second step will be activated, clicking in the "Proceeed in normal mode" will set the third step as active but the button to "Start upgrade" will be greyed out...

Now not sure if this is due to this changes or was like that before but would like to test it more to see where it comes from.

Also the second step its not set as "active" in the css for some reason.

See image attached.
landing

Itxaka:
Yeah, this also happens without this patch so its something we need to take care of, disregard my comment.

Itxaka (approved):

PR #129, opened 2017-01-26
Prechecks trust backend

skazi0:
As pointed out by @Itxaka in https://github.com/crowbar/crowbar-ui/pull/128#issuecomment-274795779 there were cases where landing page didn't seem to work properly. It turned out that this was triggered by the dummy backend returning inconsistent response (i.e. failing prechecks but `best_method` not `none`). The UI based some decisions on one part of data while others on different one. At the same time, redundant work was done in the UI instead of using the information returned from the backend.

This PR is removing re-calculation of info which is already present in the response. By this the UI should behave more consistently. Note, that the response itself can still be inconsistent: it can have failing prechecks and claim that the upgrade can be started but the idea is that the UI should trust the backend on that instead of trying to re-check things on it's own.

Note: first commit is increasing tests coverage to not go below the threshold after the actual changes in controller.

### DIFF
--- a/assets/app/features/upgrade/controllers/upgrade-landing.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-landing.controller.js
@@ -174,37 +174,26 @@
                         // Store the upgrade best method
                         vm.mode.type = response.data.best_method;
 
-                        var checksValidity = [],
-                            checksErrors = {};
+                        var checksErrors = {};
 
                         _.forEach(response.data.checks, function(check, checkKey) {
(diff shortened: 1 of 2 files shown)


Itxaka (approved):
Locally tested and fixes all issues present on the landing page. Great!

skazi0:
[Upgrade UI: Unify text on the screen about stopping Openstack services](https://trello.com/c/MHoTNz7Y/2055-upgrade-ui-unify-text-on-the-screen-about-stopping-openstack-services)

PR #132, opened 2017-01-27
upgrade: Add onFailed and postSync callbacks

skazi0:
When syncing status flags (usually on controller activation) there
can be cases where status of checked step is `failed`. Some actions
might be needed to cover this. An optional `onFailed` callback was
added for those cases.

The second (optional) added callback, `postSync` is called always
independent of the step status. This can be used to chain some action
after the status was synced as the syncing itself is asynchronous.


Itxaka (approved):
Very nice!

PR #140, opened 2017-02-09
upgrade: Fix display of errors with duplicate lines

skazi0:
By default ngRepeat requires the items to be unique. In case of error messages
some of the lines can easily be duplicated. This change switches to "track by index"
so that all lines are displayed properly.


Itxaka (approved):
Great catch

PR #144, opened 2017-02-15
upgrade: Pass response to all sync callbacks

skazi0:
Having access to response data in callbacks passed to `syncStatusFlags` is
useful in some situations. Response was already passed to onFailed. This
change adds the response argument to remainings ones. Additional benefit
is that it makes the callback calls compatible with the ones from
`waitForStepToEnd` so the same functions can be reused.


jsuchome (commented):
I'm sorry, this code is really unknown to me :-(

Itxaka (approved):

PR #153, opened 2017-03-02
upgrade: Yellow failures of non-required checks

skazi0:
Prechecks marked as not required for the upgrade are only important
for the non-disruptive mode. Their failures are not fatal so they
should not use red color as it might be misleading for the user.

zrzut ekranu z 2017-03-02 12-46-01

### DIFF
--- a/assets/app/features/upgrade/controllers/upgrade-landing.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-landing.controller.js
@@ -188,6 +188,7 @@
                             // skip unknown checks returned from backend
                             if (checkKey in vm.prechecks.checks) {
                                 vm.prechecks.checks[checkKey].status = check.passed;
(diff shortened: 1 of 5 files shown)


skazi0:
[crowbar-ui: differentiate between required and non-required prechecks](https://trello.com/c/BHL5JqmX/331-crowbar-ui-differentiate-between-required-and-non-required-prechecks)

jsuchome:
@vuntz do you want to comment? I'm fine with this

jsuchome (approved):

Itxaka:
Looks good but we shold probably mention it on the upgrade guide/docs or provide a tooltip with a text so its clearer that is not fatal. Like that, for me it suggest that its between failure and success.

Itxaka (approved):
Better than what we have

vuntz (approved):

PR #154, opened 2017-03-02
upgrade: New special error code from services step

jsuchome:
During services step, we do a check for running instances
in case of disruptive upgrade.


jsuchome:

skazi0 (approved):

PR #155, opened 2017-03-03
Increase the time before the status call during the nodes upgrade.

jsuchome:
Upgrade of nodes takes many minutes, probably even hours. We
do not need to execute the status query every 5 seconds.


Itxaka (approved):
Good to me

skazi0 (approved):

PR #158, opened 2017-03-03
upgrade: Indicate completed database step

skazi0:
The title on database configuration page changes after the step is completed
to avoid confusion (e.g. when re-opening the UI).
Also the form fields are disabled when database configuration is running or
completed to clearly indicate that this data cannot be modified anymore.

zrzut ekranu z 2017-03-03 14-18-29

### DIFF
--- a/assets/app/features/upgrade/i18n/en.json
+++ b/assets/app/features/upgrade/i18n/en.json
@@ -76,6 +76,7 @@
             },
             "database-configuration": {
                 "title": "Enter Credentials and/or Information to create or connect to a PostgreSQL service to be used for serving the Crowbar database",
+                "title-completed": "This step has been completed. Click on \"Next\" to continue the upgrade process.",
                 "navigation-tabs": {
                     "create-database": "Create a Database",
                     "connect-database": "Connect to a Database"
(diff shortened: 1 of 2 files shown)


skazi0:
[(1) S56P8 bsc#1018562 improve DB page after step was completed](https://trello.com/c/tX3JkTKK/337-1-s56p8-bsc-1018562-improve-db-page-after-step-was-completed)

assets/app/features/upgrade/i18n/en.json
```diff
@@ -76,6 +76,7 @@
             },
             "database-configuration": {
                 "title": "Enter Credentials and/or Information to create or connect to a PostgreSQL service to be used for serving the Crowbar database",
+                "title-completed": "This step has been completed. Click on button labeled \"Next\" to continue the upgrade process.",
```
jsuchome: Could we just have *Click on "Next" button...* here? Unless there's some policy for this wording...
I can be wrong of course and yours is better...

skazi0: For base version I directly took what was proposed. I didn't really like this wording either. I changed it now to better match the rest of places (which are not really 100% consistent regarding clicking (on) buttons).

jsuchome (approved):

PR #160, opened 2017-03-06
upgrade: rename non-disruptive to non_disruptive

skazi0:
Mode value was changed to symbol in the backend and could not contain
dash ('-') anymore.
In addition some missing pieces were added to the mocked backend.


jsuchome (approved):

skazi0:

PR #162, opened 2017-03-09
upgrade: Handle multi-node upgrade status

skazi0:
During disruptive upgrade multiple nodes can be upgraded in prallel.
The status structure was adapted to this on the backend side.
This change updates the UI to match the backend.

**EDIT**
Final screenshots attached below in a separate comment.

1 node:
zrzut ekranu z 2017-03-09 11-53-47
2-4 nodes:
zrzut ekranu z 2017-03-09 11-53-22
&gt;4 nodes:
zrzut ekranu z 2017-03-09 11-52-34

### DIFF
--- a/assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
@@ -32,7 +32,9 @@
 
         vm.nodesUpgrade = {
             beginUpgradeNodes: beginUpgradeNodes,
-            currentNode: undefined,
+            currentNodes: [],
+            currentAction: null,
+            maxDisplayNodes: 4,
             upgradedNodes: 0,
             totalNodes: 0,
             completed: false,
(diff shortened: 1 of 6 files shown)


skazi0:

jsuchome (approved):

Itxaka (approved):

vuntz:
I'm fine with it, but have two comments/questions:

 - can we write: "controller.1234.suse.com (role: controller)" instead of "controller.1234.suse.com / role: controller"
 - is there a way to get info on the "2 more" somehow? From the cli, maybe? I can imagine it'd be important in case things are hanging.

skazi0:
@vuntz 

- as we always group nodes by role, we can even remove the role from those lines and add it as a separate line like in single node version.
- the information is visible in the server response and in status file, I don't think we trim the output in cli so it should be visible there too.

jsuchome:
Yes, CLI output shows always everything

skazi0:
Final screenshots:
zrzut ekranu z 2017-03-09 14-07-28
zrzut ekranu z 2017-03-09 14-08-09

PR #168, opened 2017-03-20
upgrade: Display old errors when opening DB page

skazi0:
Errors from previous step runs are stored in status file. When re-opening
DB configuration page after failed run the errors from status file will now
be displayed to the user.


vmoravec (approved):
Patch tested successfully on our cloud

jsuchome (approved):

PR #175, opened 2018-05-16
upgrade: Remove CEPH related code

skazi0:
Crowbar managed CEPH clusters are not supported in Cloud7->8 upgrade.

### DIFF
--- a/assets/app/features/upgrade/upgrade.constants.js
+++ b/assets/app/features/upgrade/upgrade.constants.js
@@ -3,8 +3,7 @@
     angular
         .module('crowbarApp.upgrade')
         .constant('ADDONS_PRECHECK_MAP', {
-            'ha': ['clusters_healthy'],
-            'ceph': ['ceph_healthy']
+            'ha': ['clusters_healthy']
         })
         .constant('UNEXPECTED_ERROR_DATA', {
             title: 'unexpected_error',
--- a/assets/app/features/upgrade/controllers/upgrade-landing.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-landing.controller.js
@@ -38,10 +38,6 @@
     ) {
         var vm = this,
             optionalPrechecks = {
-                ceph_healthy: {
-                    status: false,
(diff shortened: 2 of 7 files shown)


assets/app/features/upgrade/upgrade.constants.js
```diff
@@ -3,8 +3,7 @@
     angular
         .module('crowbarApp.upgrade')
         .constant('ADDONS_PRECHECK_MAP', {
-            'ha': ['clusters_healthy'],
-            'ceph': ['ceph_healthy']
```
jsuchome, line 7: Note there there's this new group called `ceph_status` (which could contain crowbar_ceph_present error)

skazi0: yup, updating the PR right now.

jsuchome:
@skazi0 what's the branching and packaging policy here? AFAIK we need this bit on SOC7 side (but not to mess with SOC7 side of 6-7 upgrade)

skazi0:
@jsuchome re branching policy... this might get tricky. we have normal branching so far so Cloud7 has the Cloud6->7 version. I'll have to think about this...

assets/app/features/upgrade/upgrade.constants.js
skazi0: should be fixed now. see my assumptions comment below.

skazi0:
assumptions:
- no ceph addon at all (i.e. not reported from backend or ignored)
- no ceph repochecks
- no ceph health check (negative ceph presence check instead)
- anything more?

jgravel:
There's a reference of 'ceph_healthy' at https://github.com/crowbar/crowbar-ui/blob/master/routes/api/upgrade/prechecks.js#L20. Should that line be updated as well?

jsuchome (approved):

skazi0:
rebased and ready for final review

jsuchome (approved):

jgravel (approved):

PR #179, opened 2018-06-28
upgrade: Add ability to postpone upgrade

jgravel:
First draft of UI change for the postpone upgrade story. ** WIP **


assets/app/features/upgrade/i18n/en.json
```diff
@@ -109,11 +110,15 @@
             },
             "upgrade-nodes": {
                 "title": "Begin upgrading and migrating nodes by clicking on the button below",
-                "description": "Depending upon the size of your cloud and workloads this process might take quite some time.",
+                "description": "This process might take some time depending upon the size of your cloud and workloads. Uncheck the 'Upgrade Compute Nodes' box if you wish to only upgrade the controller nodes and postpone the upgrade of compute nodes. You can start the upgrade process of the compute nodes again after the controller node upgrade has been completed.",
+                "alt-description": "The Upgrade Wizard has detected that you have upgraded your controller nodes. Press the 'Upgrade Nodes' button to continue with the compute node upgrade now.",
```
jack-likes-to-code: Press the 'Upgrade Nodes' button to upgrade your compute nodes.

assets/app/features/upgrade/i18n/en.json
```diff
@@ -109,11 +110,15 @@
...
+                "upgrade-option": "Upgrade Compute Nodes",
                 "form": {
                     "upgrade-nodes": "Upgrade Nodes"
                 },
-                "finished": "Upgrade is completed. Click Finish button to go back to the dashboard.",
+                "finished": "Upgrade is completed. Click the Finish button to go back to the dashboard.",
+                "partially-upgraded": "Upgrade of the controller nodes has been completed. Click the 'Upgrade Compute Nodes' checkbox then press the 'Upgrade Nodes' button to upgrade the compute nodes now. Otherwise, exit this wizard if you wish to upgrade the compute nodes at a later time.",
```
jack-likes-to-code: checkbox, then

assets/app/features/upgrade/i18n/en.json
```diff
@@ -109,11 +110,15 @@
...
+                "resume": "Click the 'Upgrade Nodes' button to proceed with the controller node upgrade.",
```
jack-likes-to-code: to upgrade your controller nodes.

jack-likes-to-code (changes requested):
minor nitpicks

assets/app/data/crowbar/services/upgrade-status.factory.js
```diff
@@ -85,18 +85,25 @@
                             // on step failure trigger error handler
                             onError(response);
                         } else {
+                            // check for the case of postpone upgrade of compute nodes, after controller nodes were
```
skazi0: This is specific to the "nodes" step. Can you implement this inside the `onRunning()` callback on the last page and keep this service generic?

skazi0 (changes requested):

jsuchome (commented):
Few comments:

1. ￼ _Upgrade Compute Nodes_ checkbox should be disabled when the upgrade is running (just like the _Upgrade Nodes_ button

2. I can see that after controller nodes upgrade, you are disabling _Upgrade Nodes_ button unless the checkbox is checked. This makes sense, however it would be possible differently:

 2.1. Right after controller nodes upgrade, automatically check that checkbox and disable it, as no other value makes sense for it

 2.2. Enable the _Upgrade Nodes_ button also automatically (because we have the checkbox value fixed)

 2.3. Also, in this way it would be possible to have the _Upgrade Nodes_ and _Go to Dashboard_ buttons enabled at the same time - user can just choose which way to go

jgravel:
Thank you for all of the comments. As I mentioned that this PR is WIP, all the controls' enable/disable are still being worked on. However, here are my response to @jsuchome's 2nd comment:
I think automatically select the check box and enable the Upgrade Nodes button after the controller node upgrade can be confusing to the user. Since the user started the upgrade with the check box unchecked, it might not be clear as to why the check box is now checked again. If we leave the user choice intact after the controller upgrade, it confirms that their selection has provided the output as they expected. Then if the user wants to proceed with the compute node upgrade, s/he has to intentionally click on the check box. This will force the user to pay attention to the UI and will also prevent accidentally upgrading of the compute nodes when s/he is not quite ready. To me, having the Upgrade Nodes button enable/disable in reverse of the Go To Dashboard button provides 2 distinct paths to the user, which convey the message clearly that they have two simple choices to pick from at that moment.

assets/app/data/crowbar/services/upgrade-status.factory.js
jgravel: Thanks for your explanation. That makes sense. I'll make the change.

skazi0:
@jgravel I'd go more in the direction described by @jsuchome. To make things more clear maybe we should add a "green" message similar to those on the backup pages. With that, in "split upgrade" mode, after controllers are upgraded, the checkbox is checked and disabled OR hidden completely, message is displayed describing what was completed, what is the situation and what are the options to continue and "upgrade nodes" button is enabled. In current version the "decision" is explicit but still requires user to figure out by him/herself what is going on. I think clear message explaining the situation should make it all easier to understand.

**EDIT:** I have re-read the code again and I see that there are already some messages explaining the situation. Assuming that this is adjusted to the actual logic, I'd vote for no-extra-clicks-needed option.

skazi0:
@jgravel some more comments after testing:
1. "Go To Dashboard" should say "Go to Dashboard"
2. Even before any nodes are upgraded, Finish button switches to Go To Dashboard when checkbox is toggled.

I played a bit with this implementation and I'm really not sure which logic is better... the one with or without explicit checkbox after controllers are upgraded...

jgravel:
@jsuchome @skazi0 I've pushed a better version of code that should have all controls (checkbox and button) enable/disable correctly as I described in my comment above. Please take a look. In the mean time, I will ping Ken for inputs on Jiri's proposal.

gosipyan:
With help of Jacek I deployed latest changes and  went through upgrade nodes. Everything works fine,only small thing I can mention is that before starting node upgrade if I uncheck box for compute nodes upgrade "Go to Dashboard"  button appears as disabled ,don't think that it should be visible at that stage.

jsuchome:
@jack-likes-to-code are you still requesting changes in this PR?

choyj (approved):

jack-likes-to-code (approved):

jgravel:
I updated the code so that the 'Upgrade Compute Nodes' checkbox defaulted to the checked state after the controller nodes were upgraded. Please take a look.

jsuchome:
@skazi0 what about your original request, was it fixed?

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -81,6 +122,21 @@
             if (response.data.errors) {
                 return;
             }
+
+            // init upgrade compute checkbox
+            if (vm.nodesUpgrade.upgradeComputeCheckboxSelected === null) {
```
skazi0: Why not just default to `true` and override it based on the `response`?

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -81,6 +122,21 @@
...
+                if (angular.isDefined(response.data.nodes_selected_for_upgrade)) {
+                    vm.nodesUpgrade.upgradeComputeCheckboxSelected = response.data.nodes_selected_for_upgrade === 'all';
+                } else {
+                    vm.nodesUpgrade.upgradeComputeCheckboxSelected = true;
+                }
+            }
+
+            // init compute node postponed flag
+            if (vm.nodesUpgrade.computeNodesPostponedFlag === null) {
```
skazi0: Again, this should simply read the value from `response` and default to `false`.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -40,6 +46,12 @@
             completed: false,
             running: false,
             spinnerVisible: false,
+            upgradeComputeCheckboxSelected: null,
```
skazi0: I would rename it to e.g. `computeUpgradeEnabled`. Models should not refer to view terms.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -100,6 +158,33 @@
...
+            // check for the case of postpone upgrade of compute nodes, after controller nodes were
+            // upgraded, step status would still be 'running' while substep status would be 'finished'
+            var subStep = response.data.current_substep;
+            var subStepStatus = response.data.current_substep_status;
+            var nodesSelected = response.data.nodes_selected_for_upgrade
+            if (subStep === 'controller_nodes' && subStepStatus === 'finished' && nodesSelected === 'controllers') {
+                vm.nodesUpgrade.running = false;
+                // show partially upgrade message
+                vm.nodesUpgrade.upgradeControllersDone = true;
+                upgradeStepsFactory.setUpgradeStep(2);
```
skazi0: Please create constants instead of using literals numbers.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -58,11 +70,40 @@
         function beginUpgradeNodes() {
             vm.nodesUpgrade.running = true;
 
-            upgradeFactory.upgradeNodes()
+            if (vm.nodesUpgrade.upgradedNodes === 0) {
```
skazi0: This check is not correct. E.g. if something fails after first node is upgraded user should be able to continue the upgrade (re-try). This code assumes that if some node was upgraded then upgrade was postponed.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -58,11 +70,40 @@
...
+
+        function resumeUpgradeComputeNodes() {
+            vm.nodesUpgrade.resumeUpgrade = true;
+            if (vm.nodesUpgrade.computeNodesPostponedFlag) {
+                upgradeFactory.setResumeComputeNodes()
+                    .then(
+                        function() {
+                            $timeout(upgradeAllNodes, RESUME_UPGRADE_TIMEOUT_INTERVAL);
+                        },
+                        function(errorResponse) {
```
skazi0: Why not pass `upgradeError` directly?

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -100,6 +158,33 @@
...
+            var subStep = response.data.current_substep;
+            var subStepStatus = response.data.current_substep_status;
+            var nodesSelected = response.data.nodes_selected_for_upgrade
+            if (subStep === 'controller_nodes' && subStepStatus === 'finished' && nodesSelected === 'controllers') {
+                vm.nodesUpgrade.running = false;
+                // show partially upgrade message
+                vm.nodesUpgrade.upgradeControllersDone = true;
+                upgradeStepsFactory.setUpgradeStep(2);
+                if (!vm.nodesUpgrade.computeNodesPostponedFlag && !vm.nodesUpgrade.resumeUpgrade) {
+                    upgradeFactory.setPostponeComputeNodes();
```
skazi0: Why is this done in `updateModel`?

skazi0 (changes requested):

skazi0:
@jgravel generally the changes look good but i think the whole logic can be simplified and some flags/variables are redundant.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: will do

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: will do

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: will do

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: will do

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: will do

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: The obvious reason for that line being there was because it worked. For future references, if you think there's a better place for it, please don't hesitate to say so with some explanations. It would save both of us a lot of time responding back and forth between time zones. In the mean time, I will try to see where would be a better place for it.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: I've not encountered the error case that you mentioned. I don't know what the upgrade status call would return in that case. I can only code with the information I have. Will the steps.nodes.status return 'failed' while the upgraded_nodes return a value grater than 0? How about the 'current_substep_status' key? If you could provide the keys and values that I should check for, it would be helpful.

jgravel:
@skazi0 Thank you for your comments. Please see my replies inline.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
skazi0: I asked like that because it was pretty surprising to me as `updateModel` is supposed to update the variables based on some server responses. I thought that triggering postpone action would be done based on some user action (click on "go to dashboard"?).

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
skazi0: One of my test clouds is currently in such state. This is 6-7 upgrade but the responses should be pretty similar. You can check this at: http://crowbar.vh10.cloud.suse.de/upgrade/#/upgrade/upgrade-nodes

jgravel: Thanks for letting me look at your system. It's very helpful to see the upgrade status. I'll update the code.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: When I discussed with @jsuchome about how/when to call the postpone/resume API, I was under the impression that the postpone API should be called right after the controller node upgrade was done. So I coded accordingly. I can see from your point of view that it could be done based on a user action. But can you rely on the user clicking the 'Go to Dashboard' button? What if the user just moves on to crowbar ui without clicking the button, because he thinks that he could just go to the crowbar ui himself and it would be the same end result? I'm not against your idea though. If you think that we should tie the postpone API to the 'Go to Dashboard' button, I can do that.

skazi0: @jsuchome please confirm what was the intended workflow for postpone. The CLI has explicit call to postpone and resume so I thought it should be triggered by some user action in the UI as well. However it will probably not be as explicit as in the CLI.

jsuchome: I'm quite sure I wanted it to be done after explicit user action: When user clicks Go to Dashboard.
Otherwise it is not necessary - if controller nodes are upgraded, and user continues directly with upgrade of compute nodes, there's no need to postpone/resume.

It's actually written in the card: https://jira.prv.suse.net/browse/SCRD-3454 : _'Go to Dashboard' can be enabled only after successful upgrade of controller nodes. When user clicks it, API call of "nodes postpone" needs to be called so crowbar knows about the postponed state of compute nodes._

jgravel: @jsuchome I guess I didn't check that card as often as I should have. I will make the change as you specify.

jgravel:
I addressed comments from @skazi0. Please take a look.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -58,17 +73,44 @@
         function beginUpgradeNodes() {
             vm.nodesUpgrade.running = true;
 
-            upgradeFactory.upgradeNodes()
+            if (vm.nodesUpgrade.subStep === 'controller_nodes' && vm.nodesUpgrade.subStepStatus === 'finished') {
```
skazi0, line 77: Please declare those variables in `vm.nodesUpgrade`.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -81,6 +123,12 @@
             if (response.data.errors) {
                 return;
             }
+
+            // set default upgrade compute checkbox state one time only, so not to override user's choice
+            if (vm.nodesUpgrade.computeUpgradeEnabled === null) {
```
skazi0, line 128: This will still not work properly if you refresh the page after starting the upgrade. `computeUpgradeEnabled` will be reset to `null` and will switch to `true` on first model update. I think you should simply always use the `nodes_selected_for_upgrade` part of response (with no special null handling conditions).

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -40,6 +48,13 @@
             completed: false,
             running: false,
             spinnerVisible: false,
+            computeUpgradeEnabled: null,
+            upgradeControllersDone: false,
+            computeNodesPostponedFlag: false,
+            resumeUpgrade: false,
```
skazi0: This flag is not used anymore.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -58,17 +73,44 @@
...
+                upgradeFactory.upgradeNodes(vm.nodesUpgrade.computeUpgradeEnabled)
+                    .then(
+                        waitForUpgradeNodesToEnd,
+                        upgradeError
+                    );
+            }
+        }
+
+        function resumeUpgradeComputeNodes() {
+            vm.nodesUpgrade.resumeUpgrade = true;
```
skazi0: not used anymore

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -100,6 +150,35 @@
...
+            if (vm.nodesUpgrade.subStep === 'controller_nodes' && vm.nodesUpgrade.subStepStatus === 'finished'
+                && response.data.nodes_selected_for_upgrade === 'controllers') {
+                vm.nodesUpgrade.running = false;
+
+                // show partially upgrade message
+                vm.nodesUpgrade.upgradeControllersDone = true;
+                upgradeStepsFactory.setUpgradeStep(NODE_UPGRADE_STEPS.compute);
+
+                // automatically set check box to checked after upgrade controllers
+                if (!vm.nodesUpgrade.setDefaultComputeUpgrade) {
```
skazi0: I don't see why this `setDefaultComputeUpgrade` is needed. The only code guarded with that is setting of some flags. This can be done multiple times as long as the values are the same.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -100,6 +150,35 @@
...
+            // check for the case of postpone upgrade of compute nodes, after controller nodes were
+            // upgraded, step status would still be 'running' while substep status would be 'finished'
+            vm.nodesUpgrade.subStep = response.data.current_substep;
+            vm.nodesUpgrade.subStepStatus = response.data.current_substep_status;
+            if (vm.nodesUpgrade.subStep === 'controller_nodes' && vm.nodesUpgrade.subStepStatus === 'finished'
+                && response.data.nodes_selected_for_upgrade === 'controllers') {
+                vm.nodesUpgrade.running = false;
+
+                // show partially upgrade message
+                vm.nodesUpgrade.upgradeControllersDone = true;
```
skazi0: `upgradeStepsFactory.getUpgradeStep() === NODE_UPGRADE_STEP.compute` seems to match `vm.nodesUpgrade.upgradeControllersDone`. You can use the value from `upgradeStepsFactory` everywhere and remove `upgradeControllersDone`.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
```diff
@@ -90,6 +138,8 @@
                             vm.nodesUpgrade.upgradedNodes = response.data.upgraded.length;
                             vm.nodesUpgrade.totalNodes =
                                 response.data.not_upgraded.length + vm.nodesUpgrade.upgradedNodes;
+                            upgradeStepsFactory.setUpgradeAll(vm.nodesUpgrade.computeUpgradeEnabled);
+                            upgradeStepsFactory.setUpgradeStep(NODE_UPGRADE_STEPS.controller);
```
skazi0: If this was supposed to initialize the step at the beginning of upgrade then it will not work again if you refresh after nodes upgrade is started (`remaining_nodes` will not be `null` anymore). Maybe set this based on the `current_substep` from status response? This piece of code was originally meant only to set the nodes counts before upgrade is started. Rest should be available from main status response.

skazi0 (changes requested):

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: Will do.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: Will remove

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: Will remove

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: The purpose of having computeUpgradeEnabled set to null at first was because we only want to set the checkbox to true only one time at the beginning (before any node upgrade starts) in normal case (I'm not talking about the refresh case yet). The nodes_selected_for_upgrade value will not be available at that point. Also, we would be setting the computeUpgradeEnabled var every upgrade cycle, which will override the user's choice if the choice is different than the value in computeUpgradeEnabled.

Now I hadn't thought about the refresh case after starting the upgrade. I could use the nodes_selected_for_upgrade value in that case. I will look into this and make sure that both cases work.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: This code work similarly to line 126. Again, we don't want to reset the checkbox value in every upgrade cycle, specially if the user's choice is different that the value we're setting.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: vm.nodesUpgrade.upgradeControllersDone is used in upgrade-nodes-page.jade. I'm not sure how to use upgradeStepsFactory in upgrade-nodes-page.jade. If you can point me to an example of how to do that, I'd be happy to make the switch.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: Will look into this.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
skazi0: OK, I see that this is tricky case. The choice is persisted to the backend only after start of upgrade. The only way to remember user's choice across refreshes would be to use LocalStorage or something similar. I think this will be an overkill in this particular case.

skazi0: ... but still we should prefer the value from status response if it is available to properly show the state of UI on refresh after upgrade was started.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
skazi0: You can expose them via `vm` like we do here: https://github.com/crowbar/crowbar-ui/blob/master/assets/app/features/upgrade/upgrade.controller.js#L32

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
skazi0: hmmm... but `updateModel` is called only on page load and when upgrade is running. When upgrade is running user can't change these values so it's OK to sync them with the backend...

jgravel: What if the user unchecked the box and do not start the upgrade right away? For example, if he unchecks the box and gets up to get a cup of coffee. When he comes back, he'll see the box get checked back. I don't think he'll be happy about that.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: I don't see that exposing the upgradeStepsFactory call via vm will save us any time in this case because we already have upgradeControllersDone from vm. Plus upgradeControllersDone reads easier in upgrade-nodes-page.jade, comparing to getUpgradeStep() === 'compute'. But I'll change it in order to get this code closer to being merged.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
skazi0: It should not happen as `updateModel` should not be called during this period. You can add some `console.log()` to `updateModel` and watch when it is being executed. IIRC it should be on page load and periodically when upgrade is running.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
skazi0: It's not about performance but rather about keeping the state in one place. "Single source of truth" is always beneficial when you have lots of different variables. If you have two independent variables which are always changed together then in the future someone can forget to set one of them in another place which will lead to undesired results.
If you want, you can add some convenience functions to avoid the comparison. The point is that there should be only one "state" variable which holds given piece of information.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
jgravel: The upgrade cycle still happens after controller upgrade was done. This is because upgradeSuccess() was never called. If I read the code correctly, it's because the nodes' step status at this point still shows 'running', not 'passed'. The nodes' step status will change to passed only when ALL nodes were upgraded. That's why I have to account for the upgrade cycle. If this is not the behavior you're expecting, please let me know what the right behavior should be and how to get there.

jgravel:
I addressed another round of comments from @skazi0. There might be more change if @skazi0 have inputs about the upgrade cycle after controller node upgrade.

assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
skazi0: Ah, I never looked at the status responses in such scenario. I think we should extend the `waitForStepToEnd` function to allow for interruption of polling before success or error. I will prepare such change tomorrow.

skazi0: I created https://github.com/crowbar/crowbar-ui/pull/180
With this change you should be able to interrupt the polling when controllers are upgraded and computes upgrade should be postponed.

jsuchome (approved):
Seems to work as it should. Possible other fixes might come as new PR's

skazi0 (approved):
Looks good. The remaining polling problem will be fixed as followup PR.

PR #181, opened 2018-08-02
upgrade: Refactor postpone nodes upgrade

skazi0:
Code for postponing nodes upgrade was simplified and refactored.
Changes:
- status polling is stopped after controllers if partial upgrade is requested (also removed some obsolete flags)
- synchronization of `computeUpgradeEnabled` and `upgradeStepsFactory` is done using angular `$watch`
- `getUpgradeStep` is changed to `isControllersUpgraded` as it was only used for such checking
- redundant code was removed from around `beginUpgradeNodes`
- `computeUpgradeEnabled` was changed to default to true and get updates from status only when upgrade is running
- `showPartialMessage` flag was added to simplify conditions in template
- removed dependency on `upgradeFactory` from nav widget
- changed postpone to redirect to dashboard after successfull response
- renamed some constants to better match their meaning
- fixed bug: postponed message displayed when computes upgrade gets interrupted by error

### DIFF
--- a/assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
+++ b/assets/app/features/upgrade/controllers/upgrade-nodes.controller.js
@@ -16,11 +16,10 @@
         'upgradeStatusFactory',
         'upgradeStepsFactory',
         'UPGRADE_STEPS',
-        'UPGRADE_STEP_STATES',
-        'NODE_UPGRADE_STEPS',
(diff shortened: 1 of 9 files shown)


jsuchome (approved):
I do not really understand that JS code, but the UI seems to work correctly with this patch

Itxaka (approved):
It took me a while but it seems ok for me

PR #182, opened 2018-08-07
upgrade: Dummy backend for status testing

skazi0:
Added new dummy backend for UI testing. Primary usecase and main
functionality is simulation of changing upgrade status based on
responses recorded from real cloud. Example recordings are included.


Itxaka (approved):
wow, nice!

nicolasbock:
How would you use this? Run it and then interactively choose a proper response?

skazi0:
@nicolasbock yes. The main goal was to simulate "nodes" upgrade step where UI polls for status. In theory the UI should sync to the status response on load/refresh so you can jump to any page by selecting status response. Most of the POST endpoints are missing so you can't go over the whole upgrade yet but it could be extended if needed. The interactive part can also be extended to e.g. return different precheck results or error responses.

nicolasbock (approved):
Thanks, that's a really cool idea!

PR #185, opened 2019-01-11
upgrade: Fix "Next" button behavior

skazi0:
When user opened page for already completed step (e.g. via Back button)
for some steps the Next button was incorrectly enabled. This was a race between
code which enforces correct step to be displayed and code in some pages which
enabled Next button if the step was already completed.

In addition this change fixes bsc#1027306 (don't require user to re-run backup
steps).


Itxaka (approved):
Looks good!

jsuchome (approved):
I don't really understand JS code, but I trust you guys to see it is correct

PR #187, opened 2019-07-05
Pin gulp-angular-templatecache version

skazi0:
Some recent changes in gulp-angular-templatecache and/or it's deps
causes failures when building the application. Since we're already
using pretty old versions of modules here, pinning this seems like
a reasonable solution.


Itxaka (approved):
good catch!

PR #193, opened 2019-12-09
upgrade: Added missing error title

skazi0:
Title for `clusters_health_unready_nodes` was missing.

### DIFF
--- a/assets/app/features/upgrade/i18n/en.json
+++ b/assets/app/features/upgrade/i18n/en.json
@@ -191,6 +191,7 @@
             "maintenance_updates_installed": "Maintenance Updates",
             "clusters_health_crm_failures": "Some crm commands failed",
             "clusters_health_failed_actions": "Some Clusters Health actions have failed",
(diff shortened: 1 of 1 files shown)


assets/app/features/upgrade/i18n/en.json
```diff
@@ -191,6 +191,7 @@
             "maintenance_updates_installed": "Maintenance Updates",
             "clusters_health_crm_failures": "Some crm commands failed",
             "clusters_health_failed_actions": "Some Clusters Health actions have failed",
+            "clusters_health_unready_nodes": "Some nodes are not ready",
```
jsuchome: What about "Pacemaker nodes are not online" ?

So that it's clear distinction from the crowbar-readiness.... I  know the key suggests this label, but I'd rather not change the key.

skazi0: done.

jsuchome (approved):
Thanks!

